Issue #34, opened 2020-08-03
vscode Even Better TOML: Authors regex pattern for validation of pyproject.toml doesn't match email addresses surrounded by <>


tbradshaw:
When using Even Better TOML with vscode, I noticed that my author byline wasn't matching.

I had written: `authors = ["Travis Bradshaw <travis@tbradshaw.net>"]`, which created an error that reads "the value doesn't match the given pattern".

I found the regex in the source to perform a cursory check, but after pasting the regex into a couple online validators it didn't seem to parse for them.  (I think maybe this is a flavor of regex that the online validators didn't support?  The discrepancies appeared to be very minor, but the regex for matching this field is a bit of a whopper and so I didn't investigate further.)

Through just trial and error, I found that "Travis Bradshaw travis@tbradshaw.net" does work, but "Travis Bradshaw <travis@tbradshaw.net>" does not.

Checking the spec for pyproject.toml, I think that this is in error: https://www.python.org/dev/peps/pep-0621/#authors-maintainers

Thank you for your time,

-- Travis

tamasfe:
Thanks, it is just an email regex with mandatory whitespace before it, and I apparently forgot to add the `<>`.
Also currently `" some@email.com"` will be valid, but it shouldn't be.

These should be small fixes but I don't have any time for this project right now.

A PR would be very welcome if you have some time.

The regex that is releated is [here](https://github.com/tamasfe/taplo/blob/master/taplo-ide/schemas/pyproject.yaml#L77), same one is used for `maintainers`, thus it is also incorrect.

tbradshaw:
Thank you!

Issue #2, opened 2020-05-23
JSON Schema


tamasfe:
I'm planning to implement JSON schema validation, but there doesn't seem to be an actively maintained validation library that is up to date with the latest drafts.

I've started working on one, but it's rather tedious, and will take some time.

tamasfe:
Schema tracking is now done in separate issues.

PR #44, opened 2020-09-27
Update pyproject.json's poetry schema

### LINKED ISSUE #35: [BUG] Wrong Poetry schema
```toml
[tool.poetry.dependencies]
python = "^3.7"

[tool.poetry.dev-dependencies]
isort = "*"
flake8 = "*"
```
generates such an error:
```toml
the required property "python" is missing Even Better TOML
```
which is not true - `python` property is only needed in `dependencies` section.

zevisert:
Updates the bundled `pyproject.json` jsonschema that the extension uses to validate `pyproject.toml`.

Resolves #35, #34

Following from #2, Remote schema `$ref`s would be a great feature to remove having to update schemas in places like this. I'm not convinced that poetry's schema **isn't** going to change in the next long while.

### DIFF
--- a/taplo-ide/schemas/pyproject.json
+++ b/taplo-ide/schemas/pyproject.json
@@ -1,307 +1,350 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "definitions": {
-    "Poetry": {
-      "title": "Poetry",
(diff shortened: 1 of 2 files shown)


zevisert:
Ah, the fork I made a few days ago is out of date. I'll re-submit!

PR #45, opened 2020-09-29
Update pyproject.json with custom version

### LINKED ISSUE #35: [BUG] Wrong Poetry schema

zevisert:
Sorry about the noise from #44, I decided to leave out any changes to pyproject.yaml here - you can port these changes to that file if you'd like.

Resolves #35, This also (re-)addresses some of the concerns from #34.

### DIFF
--- a/taplo-ide/schemas/pyproject.json
+++ b/taplo-ide/schemas/pyproject.json
@@ -1,289 +1,355 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "definitions": {
-    "Poetry": {
-      "title": "Poetry",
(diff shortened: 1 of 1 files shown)


tamasfe:
Thank you a lot!

I'll go ahead and publish a new version of the extension with this.

The YAML is there for convenience only, as I wrote the first schema in that as it allows commenting. There's no point in converting this to it, so I'll just remove the old YAML.

PR #60, opened 2020-11-18
Bump lsp-types version to match tower-lsp

RDambrosio016:
Good afternoon, we would like to use taplo and some components from taplo's LSP to provide autocompletion for toml configuration in our linter's LSP. However, taplo uses a relatively old version of lsp-types which is incompatible with tower-lsp. This PR bumps the version of lsp-types to match tower-lsp


tamasfe:
Thanks! I wanted to do this bump eventually as well. It will break `taplo-ide`, but it shouldn't matter right now.

PR #71, opened 2020-12-08
Update lsp-types dependency

darinmorrison:
This PR just updates the `lsp-types` dependency. Needed by the rslint project.


tamasfe:
Thank you, however the next release of the core library won't depend on `lsp-types` as it seems to change a lot causing compatibility issues like this one and only the `Range` and `Position` were used from it anyway, these are replaced with types contained in the library.

I'm planning the next release once I can finally merge #62 .

PR #84, opened 2021-01-05
Dein.nvim support

kkiyama117:
# About

Add schema to support [dein.nvim](https://github.com/Shougo/dein.vim) toml support.

Some mistakes may exists and some supports are incomplete (e.x: completions for vimscript) but almost all the work is done.

## Changes
- Add dein.toml schema


tamasfe:
lgtm, thanks!

PR #85, opened 2021-01-05
Refactor and fix dein schema

kkiyama117:
# About

I'm sorry but I should change descriptions of schema for `dein.nvim`for better expression in official `taplo` website, and also found some mistakes in JSON-schema, so I've fixed them.


tamasfe:
thanks!

PR #87, opened 2021-01-30
Update assert-json-diff to 2.0

davidpdrsn:
Hey 👋 

I recently released assert-json-diff 2.0.0 which had [a few breaking changes](https://github.com/davidpdrsn/assert-json-diff/blob/main/CHANGELOG.md#200---2021-01-23).


tamasfe:
Thank you!

PR #133, opened 2021-07-04
Move wasm-bindgen usage to a cargo feature to support building for Wasm without defining exports

dsherret:
Hello! Thanks for this crate! I've been investigating a bunch of TOML parsers and found yours to be the best out of all the ones I tried. I'd like to use the parser in a Wasm build, but right now the crate adds wasm-bindgen exports, which isn't desirable when it's used internally by some other Wasm code.

After this change, the code should work the same as it currently does. I think though that eventually it should be flipped so that using wasm-bindgen is opt-in rather than opt-out.

Thanks! (Also, I've used the parser to create a code formatter... I understand that taplo has a built-in code formatter, but toml is such a simple language and I prefer full control over the output)

### DIFF
--- a/taplo/Cargo.toml
+++ b/taplo/Cargo.toml
@@ -15,9 +15,11 @@ version = "0.5.4"
 crate-type = ["cdylib", "lib"]
 
 [features]
+default = ["wasm_bindgen"]
 serde = ["serde_crate", "serde_json"]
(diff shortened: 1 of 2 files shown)


tamasfe:
Thanks! I'm glad you like it!

Regarding the formatter, I'm currently rewriting it for taplo in #131. Any input/help with that is welcome.

Issue #57, opened 2020-11-07
Properly handle oneOfs in schemas


tamasfe:
Currently all oneOfs are merged for autocompletion, instead only the relevant ones should be shown to the user.

Bludator:
I think that this not only gives merged autocomplete result but also results in fails in validation:

[With this setup](https://www.jsonschemavalidator.net/s/0UY7KCWC)

VSCode extension:
```
Expected the value to be exactly one of "test1" or "test2", but it is valid for "test1" and "test2"
```
or CLI:
```
> npx @taplo/cli lint -s test-schema.json test.toml

npx: installed 20 in 5.21s
error: failed schema validation
 --> test.toml:1:1
  | 
1 | testProp = 10
  | ^ expected exactly one schema to match, but 2 schemas matched

failure: processed 1 document with 1 error
```

tamasfe:
Yes, schemas probably need an overhaul along with #136 as well.

PR #164, opened 2021-09-07
cargo-make Makefile.toml schema

starkat99:
I put together a schema for cargo-make makefiles.

 It works fine with the taplo CLI validation, but there's some quirkiness in the vscode extension, mostly with tooltips for items, where it seems to be pulling every subschema description from `oneOf`/`anyOf` schema declarations into the parent schema, and then...making them bold and larger for some reason? Anyway, that appears to be a bug on the vscode extension side, not the schema itself afaik. It might be a good test case for it.

Screenshot of the issue:
Screenshot from 2021-09-06 23-26-32

### DIFF
--- a/schemas/makefile.toml.json
+++ b/schemas/makefile.toml.json
@@ -1,0 +1,1522 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://github.com/sagiegurari/cargo-make",
+  "title": "Makefile.toml",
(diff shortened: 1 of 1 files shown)


tamasfe:
This looks great, thanks!

The `oneOf/anyOf` schemas are not properly handled in the extension right now unfortunately (#57), because they need special care, so instead they're all just merged. I have no idea why they're bold though.

PR #177, opened 2021-10-05
Add multiple_plugins on dein schema

raa0121:
dein.vim has a new schema on  https://github.com/Shougo/dein.vim/commit/58b73f278e66b90bbb042ac532ba3d2c973c2731
This PR corresponds to that.


tamasfe:
Thanks!

Issue #142, opened 2021-07-16
RuntimeError: unreachable


LJNeon:
I was editing a `Cargo.toml` file (moving a dependency to it's own subtable) and the VSCode extension crashed with the below error. I have been unable to reproduce this at all, but figured it was worth noting.

```
..\.vscode\extensions\tamasfe.even-better-toml-0.14.2\dist\server.js:81
    throw up;
    ^

RuntimeError: unreachable
    at <anonymous>:wasm-function[2096]:0x374582
    at <anonymous>:wasm-function[2798]:0x389b66
    at <anonymous>:wasm-function[2833]:0x38a1ac
    at <anonymous>:wasm-function[2732]:0x388dfd
    at <anonymous>:wasm-function[2834]:0x38a1e2
    at <anonymous>:wasm-function[2689]:0x3882d6
    at <anonymous>:wasm-function[112]:0xe5d36
    at <anonymous>:wasm-function[102]:0xc6c6f
    at <anonymous>:wasm-function[3117]:0x38c268
    at <anonymous>:wasm-function[99]:0xbab2f
```

tamasfe:
Thanks, unfortunately stack traces after wasm-opt are not really informative. However I'm almost certain that it's related to #92, which is an incomplete feature with various edge cases not handled.

LJNeon:
I just got the error again, all I did was type `[workspace]` at the end of a `Cargo.toml` file. Yet again, I am unable to reproduce it even by typing the exact same characters.
```
..\.vscode\extensions\tamasfe.even-better-toml-0.14.2\dist\server.js:81
    throw up;
    ^

RuntimeError: unreachable
    at <anonymous>:wasm-function[2096]:0x374582
    at <anonymous>:wasm-function[2798]:0x389b66
    at <anonymous>:wasm-function[2833]:0x38a1ac
    at <anonymous>:wasm-function[2732]:0x388dfd
    at <anonymous>:wasm-function[2834]:0x38a1e2
    at <anonymous>:wasm-function[2689]:0x3882d6
    at <anonymous>:wasm-function[81]:0x64325
    at <anonymous>:wasm-function[3121]:0x38c298
    at <anonymous>:wasm-function[99]:0xbab2f
    at <anonymous>:wasm-function[3120]:0x38c28c
```

tamasfe:
weird, it's not just #92 then, I'll look into it, thanks.

PR #186, opened 2021-11-05
Avoid panic if request position is out of mapper's range

buzztaiki:
Sometimes Emacs lsp-mode sends a request containing a position parameter before synchronizing with `DidChangeTextDocument`.
At that time, the `mapper` may not be able to find the stored position and panicked with `unwrap`.
This fix is to return `Ok(None)` instead of `unwrap` to avoid this case.

Perhaps this may fix the problem like a https://github.com/golang/go/issues/48249 as well.

This PR includes commit of https://github.com/tamasfe/taplo/pull/184.

### DIFF
--- a/taplo-lsp/Cargo.toml
+++ b/taplo-lsp/Cargo.toml
@@ -21,7 +21,7 @@ futures = "0.3.5"
 hex = "0.4"
 indexmap = "1.6"
 itertools = "0.10.1"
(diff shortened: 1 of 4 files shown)


tamasfe:
LGTM, thank you! This might fix the seemingly random panics in such as reported in #142.

And sorry about the slow response.

PR #190, opened 2021-11-26
Make it possible to use !Send even in Native LSP

### LINKED ISSUE #187: Native LSP builds and Send bounds
The native LSP server builds will not compile as some futures are `!Send`. The bound is required by `tokio::spawn` to support multiple threaded executions.

The simplest fix would be to use a single-threaded executor with `spawn_local` instead, multiple threads are probably unjustified anyway for any of the workloads.

buzztaiki:
fixes #187

use `tokio::task::spawn_local` and remove `Send` requirements.

Note:
The use of multi-threaded executor is still exists in the following places:
- `create_input`
- `create_output`
- `taplo-cli`

### DIFF
--- a/lsp-async-stub/Cargo.toml
+++ b/lsp-async-stub/Cargo.toml
@@ -24,12 +24,14 @@ tracing = "0.1.29"
 default = []
 tokio-stdio = [
     "tokio",
+    "tokio/rt",
     "tokio/io-util",
     "tokio/io-std",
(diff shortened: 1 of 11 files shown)


tamasfe:
Thank you!

I was not sure about removing the `Send` support from `lsp-async-stub` itself, but it certainly makes the implementation cleaner.

PR #192, opened 2021-11-26
feat: add --check option to taplo-cli format command

### LINKED ISSUE #191: [taplo-cli] '--check' flag to 'format' command
[Someone](https://github.com/engula/engula/issues/120) (also me) would like to keep all toml files with same format, like space between fields in inline table.

Maybe we should add an option like `rustfmt` to check format issue, like [this case](https://github.com/rust-lang/rustfmt#checking-style-on-a-ci-server).

DCjanus:
close #191 

This PR is not been communicate with project maintainers, so feel free to close or push code to this branch.

Open this PR to reduce time cost, since we are living in different time zone.

### DIFF
--- a/taplo-cli/src/format.rs
+++ b/taplo-cli/src/format.rs
@@ -12,6 +12,7 @@ use taplo::formatter::{self, Options};
 pub(crate) struct FormatResult {
     pub matched_document_count: usize,
     pub excluded_document_count: usize,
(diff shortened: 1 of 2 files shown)


tamasfe:
LGTM, thanks!

I'll merge this as-is, no colors support and polishing can be done later.

PR #199, opened 2021-12-26
feat: pin clap dependency version

passchaos:


tamasfe:
Thanks! Sorry for taking this long, I'm planning to upgrade to clap3 stable as it has been released. But this should not hurt until there are new cli/lsp releases.

PR #200, opened 2022-01-08
Replace CRLF line endings with LF

Juici:
Fix issues with CRLF line endings.


tamasfe:
I have no idea how CRLFs snuck in, thanks!

PR #206, opened 2022-01-28
Update rustfmt.toml.json

Aloso:
This updates the rustfmt schema to include the latest rustfmt features (9 added options and 2 stabilized), and to add a `Documentation` link to each item.

This schema was generated with [rustfmt-schema-maker](https://github.com/Aloso/rustfmt-schema-maker) and then formatted with prettier.

The diff is unfortunately bigger than necessary. The reason is that I previously used a `HashMap` for the items, so the order was unstable. I switched to a `BTreeMap`, so they are now sorted, so the next time the schema is updated, the diff will be smaller. [Here](https://github.com/Aloso/rustfmt-schema-maker/commit/204d0d88ec4b9efe520c4f9cec6211791949f064#diff-3d952971f8702bc3d22a84e30fcbc491bdfb69bbf340eb4be2dea3c135a4782c) is a more readable diff that doesn't reorder the items.

### DIFF
--- a/schemas/rustfmt.toml.json
+++ b/schemas/rustfmt.toml.json
@@ -8,406 +8,467 @@
     "patterns": ["^(.*(/|\\\\)rustfmt\\.toml|rustfmt\\.toml)$"]
   },
   "properties": {
+    "array_width": {
+      "type": "integer",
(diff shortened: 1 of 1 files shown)


tamasfe:
LGTM, thank you!

Issue #143, opened 2021-07-21
Syntax highlight not working


sdykae:
![image](https://user-images.githubusercontent.com/50224521/126424389-6527243a-af85-4e79-b31c-8deb83feecc7.png)

iuslkae:
Vscode Theme: Monokai pro

tamasfe:
seems to be a duplicate of #78

tamasfe:
> in another order of things, what do you mean with "extended colors"? i couldn't find any feature in vscode about it.

They're not a vscode feature, they're basically just extra scopes you can set colors for.

Unfortunately I don't think there's anything I can do to _"support"_ specific themes. Even Better TOML uses the `variable` scope for the keys as it fits the default themes and somewhat makes sense, in Monokai the variables just happen to be simply white. The keys are red with Better TOML because it defines the keys as `keyword`s which they clearly aren't.

The only way to solve this properly is for the themes to support the TOML scopes defined instead.

Other than this, if you have better scope suggestions for the keys (or any other part for that matter), I'm open to them.

MixmasterFresh:
The `support.type.property-name.toml` approach seems to work very well across all of the themes I could test it on(default and community supported). I opened a pr to make the switch here: #210

tamasfe:
I believe this still needs a new release (#231), other than that this can be closed.

PR #210, opened 2022-02-02
Swaps over from `variable.key.toml` to `support.type.property-name.toml` in vscode grammar

MixmasterFresh:
This should provide a much better experience when syntax highlighting in vscode. Most(but not all) themes treat `variable.key` and `variable.key.table` the same(no highlighting for either), which makes for an inconsistent and poorly supported syntax highlighting experience. As mentioned in #143 is what the json grammar uses as well.


tamasfe:
Thank you!

PR #214, opened 2022-02-13
Schema for .gitmodules

EmilySeville7cfg:


EmilySeville7cfg:
Please help me understand why settings.json don't work as expected:

```json
{
    "evenBetterToml.schema.associations": {
        "file:///home/emilyseville7cfg/Documents/mine/schemas/taplo/test/.gitmodules": "file:///home/emilyseville7cfg/Documents/mine/schemas/taplo/schemas/gitmodules.json",
    }
}
```

I want test my schema but nothing is suggested in my `test/.gitmodules` file.

tamasfe:
Hi, thanks for the PR but I'm afraid `.gitmodules` is not valid TOML, taplo won't be able to parse it, let alone suggest completions.

Also, you don't need to account for `[` and `]` in schemas, as the TOML contents are parsed and mapped to JSON values.

The schema you submitted would be valid for TOML flies that look like this:

```toml
['[submodule "something"]']

['[submodule "something-else"]']
```

But yeah, `.gitmodules` is unfortunately not TOML but rather something INI-like, and it's unlikely that taplo would be able to parse it in the near future (or if ever), sorry.

PR #216, opened 2022-02-14
Create .replit config schema

EmilySeville7cfg:
image


EmilySeville7cfg:
It's strange but autocomplete doesn't work until I rename my test `.replit` file to `.replit.toml`.

EmilySeville7cfg:
I don't know why nothing work for `.replit` but everything is fine for `.replit.toml`. [Here is my suggestion](https://replit.canny.io/general-feedback/p/allow-specify-toml-extension-for-replit-file) for .Repl.it maintainers - allow using `.replit.toml`.

EmilySeville7cfg:
I found the solution to not to add `.toml` extension: change language mode to `TOML`. But I am not sure whether it's the best solution. @tamasfe

tamasfe:
Thanks for the PR!

I'd probably leave the `"additionalProperties": false` parts out as it will cause false errors if the schema is not updated or there are any unknown/undocumented keys. Just a thought, it's not necessary and I don't know a good solution for this. Anyways, LGTM.

> I found the solution to not to add .toml extension: change language mode to TOML.

Yes that's the only way right now, maybe we should auto-detect TOML files in the extension, but that requires the LSP to be running all the time.

schemas/replit.toml.json
```diff
@@ -0,0 +1,218 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "title": ".replit config schema",
+    "description": "https://docs.replit.com/programming-ide/configuring-repl#:~:text=Configuring%20a%20Cloned%20Repl&text=replit%20file%20and%20make%20it,and%20close%20the%20visual%20editor.",
+    "x-taplo-info": {
+      "authors": ["Emily Grace Seville (https://github.com/EmilySeville7cfg)"],
+      "patterns": ["^\\.replit$"]
```
tamasfe, line 7: I don't think this would work.

Patterns need to match an entire file URI, such as `file://home/me/some/dir/.replit`.

I'd suggest a pattern like `.*\\.replit(?:\\.toml)?$` so that it matches anything that ends with `.replit`, optionally allowing a `.toml` extension.

schemas/replit.toml.json
```diff
@@ -0,0 +1,218 @@
...
+                    "additionalProperties": false
+                }
+            },
+            "additionalProperties": false
+        },
+        "languages": {
+            "description": "Language configuration",
+            "type": "object",
+            "patternProperties": {
+                ".": {
```
tamasfe, line 189: This will match a single character, is this correct?

tamasfe (changes requested):

schemas/replit.toml.json
EmilySeville7cfg: Because regex are not anchored by default it means that language name is at least one any char. I don't know exactly valid language identifier format for Repl.it to write smth better.

tamasfe: Ah I see, my bad. In this case we could probably skip a regex match by using `additionalProperties` instead, but that would allow empty keys as well. We could also leave it as-is, up to you.

EmilySeville7cfg: I prefer not to use `additionalProperties: true` until they are permitted in the tool documentation. 😄

PR #226, opened 2022-03-14
Add support  for `.rustfmt.toml`

mathiasvr:
Add support for Rustfmt configuration file with a leading dot `.rustfmt.toml`, in addition to one without `rustfmt.toml`.


tamasfe:
I added a commit that escapes the dot, otherwise it would accept any character, other than this LGTM, thanks!

Issue #228, opened 2022-03-18
Add a schema for `.cargo/config.toml` and `rust-toolchain.toml`


ian-h-chamberlain:
Hi, the VSCode extension is super useful for discoverability in things like Cargo.toml and Makefile.toml. I would like to add a schema for `.cargo/config.toml` as well.

If someone is not already working on this, I can try to take a stab at writing the schema, but wanted to make sure I'm not repeating someone else's work by opening this issue.

There are quite a few fields allowed in `config.toml`: https://doc.rust-lang.org/cargo/reference/config.html#configuration-format
So most likely, the schema would be somewhat incomplete to start out, but I think it could still be useful. Let me know what you think!

ian-h-chamberlain:
Additionally, I would like to add one for `rust-toolchain.toml`. This is a much smaller schema and I could probably contribute this first as a way to get my feet wet writing TOML schemas (I have a local one that mostly works so far).

tamasfe:
Hi, I'm not aware of anyone working on this, and would be very much welcome!

tamasfe:
Thank you for these so far!

The tools now support the schema store and new schemas should be submitted there instead. ([docs](https://taplo.tamasfe.dev/configuration/developing-schemas.html#publishing))

ian-h-chamberlain:
@tamasfe Thanks for the update! In case there are any bugfixes / updates needed to the existing schema, should I still submit them to this repo? Or do you plan to avoid maintaining them at all in favor of the schema store?

tamasfe:
The schemas should be submitted to the schema store in all cases. I'm planning to submit the schemas I authored to the store as well (`Cargo.toml` and `pyproject.toml`), I don't think there's any benefit in storing them in this repo since I consider them public domain.

The catalog available via the website is there so that the existing schemas continue to work, but I don't plan to update it in any way.

PR #229, opened 2022-03-24
Add initial `rust-toolchain.toml` schema

ian-h-chamberlain:
First part of #228.

I am hoping to get to `.cargo/config.toml` next, but it has _lots_ of fields so may take a bit longer.

Please let me know if there's anything you'd like added or changed, I think I got most of the docs (pulled from [the rustup book](https://rust-lang.github.io/rustup/index.html)), but may have overlooked or typoed something.

### DIFF
--- a/schemas/rust-toolchain.toml.json
+++ b/schemas/rust-toolchain.toml.json
@@ -12,14 +12,14 @@
   },
   "properties": {
     "toolchain": {
-      "description": "The Rust toolchain to be used.",
+      "description": "A \"toolchain\" is a complete installation of the Rust\ncompiler (`rustc`) and related tools (like `cargo`). A toolchain\nspecification includes the release channel or version, and the host\nplatform that the toolchain runs on.",
       "type": "object",
       "minProperties": 1,
       "oneOf": [
         {
           "properties": {
             "channel": {
-              "description": "The release channel of Rust to use.",
+              "description": "Rust is released to three different \"channels\": stable, beta,\nand nightly.",
               "type": "string",
               "oneOf": [
                 {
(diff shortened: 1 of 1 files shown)


schemas/rust-toolchain.toml.json
```diff
@@ -0,0 +1,364 @@
...
+                {
+                  "$comment": "Empty schema allows for custom rustup toolchains, specific nightlies, etc."
+                }
+              ],
+              "x-taplo": {
+                "links": {
+                  "key": "https://rust-lang.github.io/rustup/concepts/channels.html"
+                },
+                "docs": {
+                  "main": "Rust is released to three different \"channels\": stable, beta,\nand nightly."
```
ian-h-chamberlain, line 41: There were a couple instances where I wasn't sure if I should put the upstream docs into `x-taplo.docs.main` or just `description`, is there a preference on this?

tamasfe: I'd say avoid taplo-specific extensions whenever possible, so prefer `description` in this case. The main reason why `x-taplo.docs.main` exists is that `x-taplo` is carried over `$ref`s in the current implementation, so it allows overriding the docs for referenced schemas, but I'm revisiting this decision in the next version as it complicates things.

tamasfe:
Thanks, looks good to me!

> Please let me know if there's anything you'd like added or changed, I think I got most of the docs (pulled from [the rustup book](https://rust-lang.github.io/rustup/index.html)), but may have overlooked or typoed something

Ufortunately I don't really have time so I'd rather not review the schema itself. So it's up to you when you think it's ready to be merged, it can always be refined later as well.

ian-h-chamberlain:
> Thanks, looks good to me!
> 
> > Please let me know if there's anything you'd like added or changed, I think I got most of the docs (pulled from [the rustup book](https://rust-lang.github.io/rustup/index.html)), but may have overlooked or typoed something
> 
> Ufortunately I don't really have time so I'd rather not review the schema itself. So it's up to you when you think it's ready to be merged, it can always be refined later as well.

Understood, I will address your comment about `x-taplo` docs and give it a once-over. Btw, it looks like there was a CI failure, but it seems unrelated to my change? Just want to verify in case I need to push some change to pass CI.
Related note:

tamasfe:
> Btw, it looks like there was a CI failure, but it seems unrelated to my change? Just want to verify in case I need to push some change to pass CI.

No, sorry, the CI has been broken for a while, I'm trying to push the new version out today and fix it along with it.

> Related note: are there any test cases I should add to actually use the new schema? Even if it's just a basic toolchain file to make sure the schema isn't broken (I've only done manual testing thus far).

The schema should be verified by a JSON schema validator, but other than that there aren't any more specific tests yet.

ian-h-chamberlain:
Ok, I think this should be good to go now. Merge at your convenience I guess?

PR #239, opened 2022-04-07
fix: follow latest dein.vim json schema

4513ECHO:
I changed about these:
- Update and fix links to help file of dein.vim
- Remove deprecated properties of plugins (`on_i`, `on_idle`)
- Add new properties of plugins (`merge_ftdetect`,  `om_cmd`, `on_lua` and more)
- Some few refactoring


tamasfe:
Thanks! Sorry for keeping this in limbo for so long, now that #243 is closed, this should be good to go.

4513ECHO:
OK, Thanks a lot!

PR #247, opened 2022-04-21
Fix website CD by fixing a reference to the taplo cargo dependency

jsierles:


tamasfe:
Didn't think this would be the only issue, awesome, thanks!

jsierles:
Woops - looks like the website build is still failing. I'll look into that.

tamasfe:
Thanks, I fixed it, it was still a dependency issue, I just pinned the previous taplo version for the time being.

PR #250, opened 2022-04-30
Add `--stdin-filepath` option to format cmd

lffg:
This PR introduces the `--stdin-filepath` option to the `format` command.

This was inspired by ESLint's `--stdin-filename` and Prettier's [`--stdin-filepath`](https://prettier.io/docs/en/cli.html#--stdin-filepath). It doesn't have any behavior in the actual parsing of files or their resolution.

It is mainly useful when showing error information. In such cases, instead of getting:

```txt
  ┌─ -:2:5
  │
2 │ [bar=6
  │     ^ expected "]"
```

One would get:

```txt
  ┌─ ./foo.toml:2:5
  │
2 │ [bar=6
  │     ^ expected "]"
```
By executing `format` as `taplo format --stdin-filepath ./foo.toml -`.

### DIFF
--- a/crates/taplo-cli/src/commands/format.rs
+++ b/crates/taplo-cli/src/commands/format.rs
@@ -22,11 +22,12 @@ impl<E: Environment> Taplo<E> {
         self.env.stdin().read_to_string(&mut source).await?;
 
         let config = self.load_config(&cmd.general).await?;
+        let display_path = cmd.stdin_filepath.as_deref().unwrap_or(&"-");
 
         let p = parser::parse(&source);
 
         if !p.errors.is_empty() {
-            self.print_parse_errors(&SimpleFile::new("-", source.as_str()), &p.errors)
+            self.print_parse_errors(&SimpleFile::new(display_path, source.as_str()), &p.errors)
                 .await?;
 
             if !cmd.force {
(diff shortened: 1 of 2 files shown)


crates/taplo-cli/src/commands/format.rs
```diff
@@ -61,6 +62,10 @@ impl<E: Environment> Taplo<E> {
 
     #[tracing::instrument(level = "debug", skip_all)]
     async fn format_files(&mut self, cmd: FormatCommand) -> Result<(), anyhow::Error> {
+        if cmd.stdin_filepath.is_some() {
+            bail!("`--stdin-filepath` can only be used with stdin");
```
tamasfe, line 66: Bailing here feels a bit too much, the flag just rewrites `-` to be whatever, maybe we could just ignore it here or probably print a warning?

tamasfe (changes requested):

tamasfe:
Thanks, I left a comment about a tiny detail, otherwise LGTM.

crates/taplo-cli/src/commands/format.rs
lffg: Yes, I totally agree with you here. Thanks.

tamasfe (approved):

PR #257, opened 2022-05-29
Fix broken image links due to a previous file reorganization

kkaja123:
This only affected the vscode readme. Image files were placed in the `images` sub-directory for the vscode directory, but its corresponding `README.md` was not updated with the new paths.


tamasfe:
Thanks!

PR #273, opened 2022-06-20
Don't auto close string delimiters in strings.

sebthom:
Also don't auto close apostrophe in comments

### DIFF
--- a/editors/vscode/language-configuration.json
+++ b/editors/vscode/language-configuration.json
@@ -10,22 +10,10 @@
 		]
 	],
 	"autoClosingPairs": [
-		[
-			"\"",
-			"\""
-		],
-		[
-			"'",
-			"'"
-		],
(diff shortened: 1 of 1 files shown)


sebthom:
Basically this, which is a comment from another user for a related issue in another textmate project:
"When I'm writing a comment or a string, I don't want a pair to be inserted when typing characters such as apostrophes for instance, especially as this moves my cursor around. Similarly, when I'm in the middle of writing a string but for some reason I have deleted the end delimiter, I want the string to be closed with a unique delimiter character rather than a pair being inserted. The current behaviour is often frustrating."

Especially for apostrophes inside strings and comments (which from the grammar's pov is unstructured data) it usually used for possessive nouns.

tamasfe:
Fair points, thanks, LGTM.

PR #287, opened 2022-07-22
pyproject.toml.json to match poetry author regex

laundmo:
Using the same regex used by Poetry, with the named capturing groups removed. Main difference is that the email in poetry is optional.

Poetry defines its AUTHOR_REGEX here:
https://github.com/python-poetry/poetry-core/blob/cd444099391eaf237fbe2aea0e82f5e296f06b1b/src/poetry/core/packages/package.py#L33


tamasfe:
Thank you, but I'm not accepting schema changes anymore since they are likely to be migrated to the schema store soon.

Sorry for not being clear about it, I've just made an issue regarding this. #288

PR #301, opened 2022-08-15
add delta view to taplo-cli

drahnr:
It's not uncommon to use a formatter in CI where a subset of contributors do not have the tool installed. That set of contributors is usually very appreciative of direct cues what to fix.

The changes:

Adds a dependency to `prettydiff` and `ansi_term` for creating the diff and coloring, there is currently (almost) no clustering of diff lines, this is to be improved.

### DIFF
--- a/crates/taplo-cli/src/commands/format.rs
+++ b/crates/taplo-cli/src/commands/format.rs
@@ -66,11 +66,19 @@ impl<E: Environment> Taplo<E> {
         Ok(())
     }
 
-    fn print_diff(path: impl AsRef<Path>, original: &str, formatted: &str) {
+    async fn print_diff(&self, path: impl AsRef<Path>, original: &str, formatted: &str) {
         let path = path.as_ref();
-        println!("diff a/{path} b/{path}", path = path.display());
-        println!("--- a/{path}", path = path.display());
-        println!("+++ b/{path}", path = path.display());
+
+        // print to stdout
+        macro_rules! echo {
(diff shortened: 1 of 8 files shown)


crates/taplo-cli/src/commands/format.rs
```diff
@@ -65,6 +66,92 @@ impl<E: Environment> Taplo<E> {
         Ok(())
     }
 
+    fn print_diff(path: impl AsRef<Path>, original: &str, formatted: &str) {
+        let path = path.as_ref();
+        println!("diff a/{path} b/{path}", path = path.display());
```
tamasfe, line 71: The stdio is behind the `Environment` abstraction in order to support WASM. The standard output should be available here via `self.env.stdout()`, all printing should be done via that instead.

tamasfe (changes requested):

tamasfe:
I think this could be gated behind a flag like `--diff` as this can get very verbose, looks good otherwise, thanks!

crates/taplo-cli/src/commands/format.rs
drahnr: Thanks, I wasn't aware

drahnr:
Added `--diff` flag, both for `--check` and non-check. Introduced a scope macro `echo!` that utilizes `self.env`. I think it's ready for another review pass :)

tamasfe:
Looks great, thank you!

PR #302, opened 2022-08-16
drop the openssl dependency

drahnr:
Removes the dependency on openssl by using the `rustls-tls` feature flags across all toplo packages.


tamasfe:
LGTM, thanks!

PR #327, opened 2022-10-05
fix(cli): remove build script

zertosh:
This env's use was removed in #211, so the build script is unnecessary.


tamasfe:
Thanks! I don't even remember what it was used for.

PR #334, opened 2022-10-13
feat: allow to build with native-tls instead of rustls

jirutka:


tamasfe:
Sorry for the delay, the feature is pretty reasonable, thanks!

Issue #320, opened 2022-09-21
VS Code: this document has been excluded


muuvmuuv:
Cannot find a way to include a dotfile written in TOML. VS Code always prints "this document has been excluded". Adding a config file with include "**/.file" does not work. File associations neither. Anything I have missed?

alkatar21:
I have the same message for Python `Pipfile` with the VSCode extension. Is there any extension/VSCode setting im missing to add Pipefile as a valid toml file?

gutenfries:
+1 same issue

reinderien:
I have this issue on Even Better TOML 0.19.2. Try as I might, `settings.json`

```js
    "files.associations": {
        "**/*.wara": "toml"
    },
```

has no effect and the angry squiggle always says _this document has been excluded_.

Rudxain:
Has anyone tried [this](https://github.com/tamasfe/taplo/issues/580#issuecomment-2004174721)?

senekor:
Works for me in helix. I can also [confirm](https://github.com/tamasfe/taplo/issues/580#issuecomment-2398796949) that the CWD of helix must be a parent of the file, which is good enough for me. To save you a click, add the following to your `~/.config/helix/languages.toml`:

```toml
[[language]]
name = "toml"
roots = ["."]
```

liblaf:
I solved this issue by adding a `.taplo.toml` under my `$HOME` directory:

```toml
# file: $HOME/.taplo.toml
#:schema https://json.schemastore.org/taplo.json

include = ["**/*.toml", "**/*.toml.tmpl"]
```

PR #378, opened 2023-02-03
fix: language server include rule matching

sethp:
Previously, using a configuration like:

```taplo.toml
[formatting]
reorder_keys = false
align_entries = true

[[rule]]
include = ["**/taplo.toml", "**/.taplo.toml"]

[rule.formatting]
reorder_keys = true
```

Running `taplo fmt` would  sort the two formatting keys, but the language server (i.e. "format document" in vscode) would not.

The trouble is that the language server was comparing an include rule (prefixed with the absolute path of a workspace) against a file URI directly, and a GlobRule that's including
`/absolute/path/**/taplo.toml` does not match a normalized URI string like `file:///absolute/path/taplo.toml`.

Also, because of the path prefix, changing the config file to

```
include = ["file://**/taplo.toml"]
```

produces a similar result (neither `taplo fmt` nor the server apply the rule), because `file:///abs/path/taplo.toml` also does not match the glob `/abs/path/file://**/taplo.toml`.

### DIFF
--- a/crates/taplo-lsp/src/handlers/formatting.rs
+++ b/crates/taplo-lsp/src/handlers/formatting.rs
@@ -1,8 +1,7 @@
 use lsp_async_stub::{rpc::Error, util::LspExt, Context, Params};
 use lsp_types::{DocumentFormattingParams, TextEdit};
-use std::path::PathBuf;
 use taplo::formatter;
-use taplo_common::{environment::Environment, util::Normalize};
+use taplo_common::environment::Environment;
 
 use crate::World;
 
@@ -23,7 +22,15 @@ pub(crate) async fn format<E: Environment>(
         }
     };
 
-    let doc_path = PathBuf::from(p.text_document.uri.as_str()).normalize();
+    let doc_path = context
+        .env
+        .to_file_path_normalized(&p.text_document.uri)
+        .ok_or_else(|| {
+            Error::invalid_request().with_data(format!(
+                "invalid (non-local) uri for file: {}",
+                p.text_document.uri
+            ))
+        })?;
 
     let mut format_opts = formatter::Options {
         indent_string: if p.options.insert_spaces {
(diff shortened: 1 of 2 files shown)


sethp:
I suspect this is at least related to #320 , too.

crates/taplo-lsp/src/handlers/formatting.rs
```diff
@@ -43,6 +50,15 @@ pub(crate) async fn format<E: Environment>(
...
         .update_format_options(&doc_path, &mut format_opts);
 
+    let scopes = ws.taplo_config.format_scopes(&doc_path).collect::<Vec<_>>();
+    tracing::trace!(
+        ?doc_path,
+        ?format_opts,
+        ?scopes,
+        all_rules = ?ws.taplo_config.rule,
+        matched_rules = ?ws.taplo_config.rules_for(&doc_path).collect::<Vec<_>>(),
+    );
```
ia0, line 60: I'm assuming this is for debugging. Unless you have a strong opinion about it, I would suggest keeping `ws.taplo_config.format_scopes(&doc_path)` inlined below (as it was before) and not round-trip through a `Vec` for the sake of tracing at trace level.

ia0 (commented):
Thanks a lot @sethp for the fix!

I could reproduce the issue in Emacs and I confirm that your change is fixing the issue. I left a comment though that I think should be addressed before merging if that's fine with you.

crates/taplo-lsp/src/handlers/formatting.rs
sethp: Hey @ia0 thanks for the review! I do remember the scopes feeling somewhat important to preserve so they show up in the LSP output tab with e.g. a vscode config setting like:

```json
  "evenBetterToml.taplo.environment": {
    "RUST_LOG": "trace"
  }
```

But, you'll have to forgive me if I'm not able to recall the precise problem I thought I was solving. 

Either way, I hear you about wanting to avoid the round trip unless we're going to be tracing: what do you think of 6031848 ?

ia0: Thanks! This looks good to me.

ia0 (approved):

ia0:
@panekj Could you take a look at this bug fix?

panekj (approved):

PR #379, opened 2023-02-03
docs: document Trunk Check integration

sxlijin:
Trunk Check is a universal linter which integrates with a wide variety of linters and formatters, `taplo` included.

We're big fans of `taplo` and figured that you might find our tool to be interesting enough to include it here.

---

`taplo` has been super useful for us, and we've had a few users [ask us about our integration with you](https://trunkcommunity.slack.com/archives/C0205B6KK8X/p1674743406903219), so we figured that it would be nice to also add documentation from your side :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,13 @@
 
 This is the repository for Taplo, a TOML v1.0.0 toolkit, more details on the [website](https://taplo.tamasfe.dev).
 
-
 - [Taplo](#taplo)
   - [Status](#status)
   - [Contributing](#contributing)
+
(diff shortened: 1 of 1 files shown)


ia0:
Thanks @sxlijin for your interest and support!

Trunk Check is an interesting effort, but I don't think the README of taplo is the right place to advertise it. Taplo is supposed to integrate well with many things, so giving a list of integrations will not communicate the right message. I'm thus closing this PR for the time being. Feel free to reopen if you have another perspective or suggestion.

PR #354, opened 2022-11-20
Handle LSP shutdown request

the-mikedavis:
These lines in the Server loop discard any messages with the shutdown method. The Server can handle and correctly respond to shutdown requests, though, so there's no need to discard the message.

PR #401, opened 2023-04-29
Handle shutdown requests

### LINKED ISSUE #400: The LSP server doesn't reply to shutdown
I'm trying to use taplo with eglot in emacs. When I shutdown the LSP server (with eglot-shutdown), it doesn't reply. Here's the relevant log:

```
[client-request] (id:3) Sat Apr 29 13:29:57 2023:
(:jsonrpc "2.0" :id 3 :method "shutdown" :params nil)
```

The LSP server doesn't reply to this request. I'm able to fix the problem with #401.

ia0:
Fixes #400

### DIFF
--- a/crates/lsp-async-stub/src/listen.rs
+++ b/crates/lsp-async-stub/src/listen.rs
@@ -64,8 +64,6 @@ impl<W: Clone + 'static> Server<W> {
                                     drop(output);
                                     drop(input);
                                     break;
-                                } else if msg.method.as_ref().map(|m| m == "shutdown").unwrap_or(false) {
-                                    continue;
                                 }
 
                                 let task_fut = self.handle_message(


irevoire:
Shouldn’t we match the shutdown request here as well to drop the input/output?
```rust
if msg.method.as_ref().map(|m| m == "exit").unwrap_or(false) {
                                    drop(output);
                                    drop(input);
                                    break;
                                }
```

ia0:
> Genuine question; shouldn’t we match the shutdown request here as well to drop the input/output?

I don't know about the input/output dropping but we definitely don't want to exit the loop. We want to execute the `handle_request` method.

Philipp-M:
Duplicate of #354

ia0:
> Duplicate of #354

Very good point! How did I miss that one? I'm going to contribute there and closing this.

Issue #403, opened 2023-04-30
Looking for maintainers


tamasfe:
I had many plans with this project but it has become apparent that I currently cannot seem to dedicate any time to OSS and it's not clear when this will change.

I'm looking for maintainers to take over this repo, mainly to triage PRs and publish new versions.

panekj:
I've been watching over the issues and PRs for some time now but I haven't been able to act on anything yet since I'm spread thin over different projects but I would like to help out and should be able to focus a little bit more on this project fairly soon
(My motivation is proper TOML support for https://github.com/lapce/lapce)

ia0:
I don't want to commit just yet, but I'd like to co-maintain. Taplo is an important tool in my workflow providing considerable value. I'd like to keep it in good state.

tamasfe:
Thank you @panekj and @ia0. I have invited you both as collaborators.

I'll somehow have to renew the vscode marketplace token, other than that releases should _just work_ via github actions, so they are not blocked by my absence anymore.

ia0:
Thanks @tamasfe !

@panekj How do you want to go through the backlog? I'd suggest those priorities:
1. Going through the continuous integration and make sure there is enough testing that we could be confident changing things without breaking too much. (For example, I didn't see much testing about LSP interactions.) Or at least be aware of which areas need particular attention because they are not as well tested as others.
2. Going through the pull requests: triaging and starting to merge the most simple ones (like dependabot PRs, which should also reenable it automatically because it's currently paused).
3. Going through the issues: mostly triaging.
4. Make a release if needed.

panekj:
I would like to see https://github.com/tamasfe/taplo/pull/361 and https://github.com/tamasfe/taplo/pull/366 merged first and rest is best effort :)

ia0:
> I would like to see #361 and #366 merged first and rest is best effort :)

Sounds good. I've reviewed those 2 PRs. On my side, I'd like at least to merge all dependabot PRs and #354. Should I wait for your review on them or can I go ahead and merge if I'm satisfied with the PR state?

panekj:
I think that the 2nd person that gives approval can merge them. I would like it to be coordinated and prevent any mistakes, so IMO, PRs should be reviewed by both of us, then someone can merge them.

ia0:
I would also suggest another convention for merging PRs:
1. If the PR is not authored by us, then it should be approved by both of us and the second approver merges it.
[...]

PR #409, opened 2023-05-18
Add `align_single_comments` option

### LINKED ISSUE #404: Smarter "align_comments" when there's only 1 comment?
:wave: Howdy!

We've added support for Taplo to https://github.com/pantsbuild/pants, and are not integrating it into the repo itself.

One thing I noticed was I like the idea of "align_comments" (:+1:) but I wished that if there was 1 (and only 1) comment, it didn't apply.

E.g.
```toml
foo = [
    "hello world",
    "a",                # I'm way over here
    "b",
]
```
would be:
```toml
foo = [
    "hello world",
    "a",  # I'm way over here
    "b",
]
```

but adding a new comment would result in
```toml
foo = [
    "hello world",
    "a",                # I'm way over here
    "b",                # me too
]
```

This, of course, would be hidden behind a toggle.

Thanks again (I'm also looking forward to integrating Taplo into our eventual Pants LSP for handling our config file, so thanks for the LSP side of things too :raised_hands: )

thejcannon:
Fixes #404 by adding a new option (as to not disrupt existing behavior) that controls whether single comments get aligned.

### DIFF
--- a/crates/taplo/src/formatter/mod.rs
+++ b/crates/taplo/src/formatter/mod.rs
@@ -182,6 +182,10 @@ impl Options {
     fn newlines(&self, count: usize) -> impl Iterator<Item = &'static str> {
         repeat(self.newline()).take(usize::min(count, self.allowed_blank_lines + 1))
     }
+
+    fn should_align_comments(&self, comment_count: usize) -> bool {
+        (comment_count != 1 || self.align_single_comments) && self.align_comments
+    }
 }
 
 #[derive(Debug, Clone)]
@@ -722,8 +726,7 @@ fn add_entries(
         })
         .collect::<Vec<_>>();
 
-    let align_comments =
-        (comment_count != 1 || options.align_single_comments) && options.align_comments;
+    let align_comments = options.should_align_comments(comment_count);
     *formatted += &format_rows(
         if !options.align_entries && !align_comments {
             0..0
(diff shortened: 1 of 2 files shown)


crates/taplo/src/formatter/mod.rs
```diff
@@ -709,18 +715,21 @@ fn add_entries(
...
                 row.push(c);
+                comment_count += 1;
             }
 
             row
         })
         .collect::<Vec<_>>();
 
+    let align_comments =
+        (comment_count != 1 || options.align_single_comments) && options.align_comments;
```
ia0, line 726: nit: We should create a `fn should_align_comments(&self, comment_count: usize) -> bool` helper function in the `impl Options` line 168 of `src/formatter/mod.rs` because we use this logic both here and below.

ia0 (changes requested):
This looks good to me modulo the small nit. Thanks for pulling this up!

crates/taplo/src/formatter/mod.rs
thejcannon: I considered it. Glad to know my intuition was aligned with yours :smile:

thejcannon:
@ia0 I can't click the big green button, so when you're ready :rocket:

ia0:
> @ia0 I can't click the big green button, so when you're ready rocket

I'm ready, but we're taking a 2 approval strategy (second approver merges), see https://github.com/tamasfe/taplo/issues/403#issuecomment-1549227674. The second reviewer is @panekj. It might take some time because we are new maintainers so even simple PRs may take time to review.

panekj (approved):

PR #448, opened 2023-08-08
add package and dependencies property for cargo.toml

ShuiRuTian:
Add two property "workspace.dependencies" and "workspace.package".

The description is got from "https://github.com/rust-lang/cargo/blob/211fd7eac105f0960433a935ce17b8d787233692/src/doc/src/reference/workspaces.md?plain=1", and replace `\n` and `\"`, kind of stupid.

### DIFF
--- a/site/site/public/schemas/Cargo.toml.json
+++ b/site/site/public/schemas/Cargo.toml.json
@@ -1297,6 +1297,19 @@
             }
           }
         },
+        "dependencies":{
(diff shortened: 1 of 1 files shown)


ShuiRuTian:
I found this: https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/cargo.json

Obviously it's a fork of this file, maybe we could take something back?

ia0:
Thanks for your PR! However this file is not used anymore and only kept for documentation on the website. Taplo now uses JSON Schema Store as you referenced in your second post. I'm closing the PR as WAI (working as intended).

PR #449, opened 2023-08-08
Update proc-macro2 to fix compilation error with nightly

ia0:
See https://github.com/rust-lang/rust/issues/113152 for more details.

This commit was generated by:

```
find . -name Cargo.toml -execdir cargo update -p proc-macro2 \;
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1654,9 +1654,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.47"
+version = "1.0.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


panekj (approved):

wmmc88:
@ia0 @panekj `cargo install taplo-cli --locked` still fails rn on nightly since no release has been made since this pr. could you do a release of taplo so that this fix is released?

ia0:
Yes I'm also hitting this issue and having to work around it. And it's even more problematic since `json_value_merge` released a breaking change in 1.2.0 (https://github.com/jmfiaschi/json_value_merge/issues/18). I would suggest bumping that to 2.0.0 first then doing a release. I've create #473 to track that.

PR #454, opened 2023-08-10
Support adding custom CA

### LINKED ISSUE #443: Add support for custom root certificate
# Request
Add an ability to define a custom root certificate for querying a schema from the Internet.

# Motivation
In my environment, it's necessary to use a custom root certificate.
I am using Taplo as a language server with Neovim. Here is what I would get when I opened a `Cargo.toml` file.
An error from Neovim `:LspLog`
```
taplo:configuration_change:initialize: failed to fetch catalog error=error sending request for URL 
(https://www.schemastore.org/api/json/catalog.json): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer\n\nCaused by:
    0: error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
    1: invalid peer certificate contents: invalid peer certificate: UnknownIssuer 
     self.root=file:///...
    OR taplo:configuration_change:initialize: failed to add schemas from catalog error=error sending request for URL 
(https://www.schemastore.org/api/json/catalog.json): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer self.root=file:///...
"
```

I have already setup my OS to trust the custom certificate. Also it seems that this crate uses `reqwest` to do the query.
I believe `reqwest` doesn't get the list of root certificates from the OS.

zynaxsoft:
Resolve #443 

Inspired by [Node.js setting](https://nodejs.org/api/cli.html#node_extra_ca_certsfile), I named the env var to be `TAPLO_EXTRA_CA_CERTS` (plural here because one file could contain many certs).

I extracted the getting `reqwest` client part as a function though I don't know exactly where the best location for this function is. I temporarily put it in `taplo-common/util.rs`.

I tested the build on my local environment described in the issue and it worked perfectly.

to: @ia0 
How should we document this?
I was thinking about making the error more obvious and adding a suggestion to let the user know they need to add the environment variable. Though I don't really know how to match the `UnknownIssuer` error from `reqwest`.

### DIFF
--- a/crates/taplo-common/src/util.rs
+++ b/crates/taplo-common/src/util.rs
@@ -120,3 +120,27 @@ pub(crate) fn normalize_str(s: &str) -> Cow<str> {
         percent_decoded
     }
 }
+
+#[tracing::instrument]
+pub fn get_reqwest_client(timeout: std::time::Duration) -> Result<reqwest::Client, reqwest::Error> {
+    fn get_cert() -> Result<reqwest::Certificate, anyhow::Error> {
+        let path = std::env::var("TAPLO_EXTRA_CA_CERTS")?;
+        let path = Path::new(&path);
+        let ext = path.extension().and_then(|ext| ext.to_str());
+        let buf = std::fs::read(path)?;
+        tracing::info!(
+            "Found a custom CA {}. Reading the CA...",
+            path.to_string_lossy()
+        );
+        match ext {
+            Some("der") => Ok(reqwest::Certificate::from_der(&buf)?),
+            _ => Ok(reqwest::Certificate::from_pem(&buf)?),
+        }
+    }
+    let mut builder = reqwest::Client::builder().timeout(timeout);
+    if let Ok(cert) = get_cert() {
+        builder = builder.add_root_certificate(cert);
+        tracing::info!("Added the custom CA");
+    }
+    builder.build()
+}
--- a/crates/taplo-common/src/util.rs
+++ b/crates/taplo-common/src/util.rs
@@ -121,6 +121,7 @@ pub(crate) fn normalize_str(s: &str) -> Cow<str> {
     }
 }
 
+#[cfg(not(target_arch = "wasm32"))]
 #[tracing::instrument]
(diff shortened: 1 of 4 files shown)


crates/taplo-common/src/util.rs
```diff
@@ -120,3 +120,27 @@ pub(crate) fn normalize_str(s: &str) -> Cow<str> {
...
+        match ext {
+            Some("der") => Ok(reqwest::Certificate::from_der(&buf)?),
+            _ => Ok(reqwest::Certificate::from_pem(&buf)?),
+        }
+    }
+    let mut builder = reqwest::Client::builder().timeout(timeout);
+    if let Ok(cert) = get_cert() {
+        builder = builder.add_root_certificate(cert);
+        tracing::info!("Added the custom CA");
+    }
```
ia0, line 144: This could also be something like:

```rust
if let Some(path) = std::env::var_os("TAPLO_EXTRA_CA_CERTS") {
    match get_cert(&path) {
        Ok(cert) => {
            builder = builder.add_root_certificate(cert);
            tracing::info!(%path, "Added the custom CA");
        }
        Err(err) => tracing::error!(error = %err, "Could not parse the custom CA");
    }
}
```

ia0 (commented):
Thanks a lot for the PR! It looks good to me.

Regarding where to put the reqwest builder function, I think that's the right place.

Regarding documentation, I would say `site/site/configuration/using-schemas.md` might be appropriate since it seems this is mostly useful for the schema catalog download. A sentence at the end of the page saying that the root certificate is customizable with the `TAPLO_EXTRA_CA_CERTS` environment variable for both the LSP and CLI is probably enough.

Regarding the error, I think using tracing should be enough like one of my comment shows. At least in Emacs it's easy to access the stderr of the LSP server. And when using the CLI it's directly visible. We could in theory improve the LSP case by producing a diagnostic or something, but that's probably overkill.

Thanks again for your work!

zynaxsoft:
I love your suggestions and I have modified the code according to your comments!

ia0 (approved):
Thanks!

panekj (approved):

Issue #434, opened 2023-07-12
LSP Server Crashing on VS Code


cs-clarence:
I get this error when using Even Better TOML VS Code pluggin, it's making the extension unusable.
This error appears when selecting a schema (it doesn't auto-detect schemas for some reason).

```bash
panicked at 'getrandom::getrandom() failed.: Error { internal_code: 2147483660, description: "Node.js crypto module is unavailable" }', /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ahash-0.8.3/src/random_state.rs:81:51

Stack:

Error
    at PA.i.wbg.__wbg_new_693216e109162396 (/home/rencedm112/.vscode/extensions/tamasfe.even-better-toml-0.19.1/dist/server.js:2:4646)
    at wasm://wasm/00fb2f8a:wasm-function[2518]:0x2e8066
    at wasm://wasm/00fb2f8a:wasm-function[4393]:0x328a27
    at wasm://wasm/00fb2f8a:wasm-function[3073]:0x303105
    at wasm://wasm/00fb2f8a:wasm-function[1499]:0x29f415
    at wasm://wasm/00fb2f8a:wasm-function[1019]:0x263047
    at wasm://wasm/00fb2f8a:wasm-function[274]:0x155b0c
    at wasm://wasm/00fb2f8a:wasm-function[362]:0x1910fd
    at wasm://wasm/00fb2f8a:wasm-function[3286]:0x30a999
    at wasm://wasm/00fb2f8a:wasm-function[889]:0x23ea21

--- snip ---

RuntimeError: unreachable
    at wasm://wasm/00fb2f8a:wasm-function[2518]:0x2e8128
    at wasm://wasm/00fb2f8a:wasm-function[4393]:0x328a27
    at wasm://wasm/00fb2f8a:wasm-function[3073]:0x303105
    at wasm://wasm/00fb2f8a:wasm-function[1499]:0x29f415
    at wasm://wasm/00fb2f8a:wasm-function[1019]:0x263047
    at wasm://wasm/00fb2f8a:wasm-function[274]:0x155b0c
    at wasm://wasm/00fb2f8a:wasm-function[362]:0x1910fd
    at wasm://wasm/00fb2f8a:wasm-function[3286]:0x30a999
    at wasm://wasm/00fb2f8a:wasm-function[889]:0x23ea21
    at wasm://wasm/00fb2f8a:wasm-function[1409]:0x295ce4
[...]

PR #456, opened 2023-08-11
Fix comments with brackets

### LINKED ISSUE #455: Formatter destoys data
Sorry for the dramatic title. `taplo fmt` does this to my array:

```diff
 ignore = [
   # __magic__ methods are are often self-explanatory, allow missing docstrings
   "D105",
-  # first line should end with a period [Bug: doesn't work with single-line docstrings]
-  "D400",
+  # first line should end with a period [Bug: doesn't work with single-line docstrings]  "D400",
   # First line should be in imperative mood; try rephrasing
   "D401",
 ]
```

This removes the `"D400"` entry from it

flying-sheep:
Fixes #455

### DIFF
--- a/crates/taplo/src/formatter/mod.rs
+++ b/crates/taplo/src/formatter/mod.rs
@@ -1106,10 +1106,8 @@ fn format_array(node: SyntaxNode, options: &Options, context: &Context) -> impl 
                         skip_newlines = 0;
                     }
 
-                    if !formatted.ends_with(']') {
-                        formatted
-                            .extend(options.newlines(newline_count.saturating_sub(skip_newlines)));
-                    }
+                    formatted
+                        .extend(options.newlines(newline_count.saturating_sub(skip_newlines)));
                 }
                 COMMENT => {
                     let newline_before = t
--- a/crates/taplo/src/tests/formatter.rs
+++ b/crates/taplo/src/tests/formatter.rs
@@ -1108,3 +1108,26 @@ my_array = [
 
     assert_format!(expected, &formatted);
 }
+
+#[test]
+fn test_comment_with_brackets() {
+    let src = r#"
+my_array = [
+  # [x]
+  "y",
+]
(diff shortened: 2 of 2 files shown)


ia0:
There's actually an easy way to fix this, but I don't know if there are other consequences.

Just remove this `if` and always execute its content: https://github.com/tamasfe/taplo/blob/5bd96253075e42dfbebe9c441611c36412c3221a/crates/taplo/src/formatter/mod.rs#L1109-L1112

This `if` looks like a hack to detect if an array just finished, but the closing bracket could be part of a comment. I don't know the situations when this `if` is false. It would be nice to figure out and add a test. If it can never happen then we're good. If it can happen, then we need to introduce a new variable to remember if we just closed a bracket outside a comment (probably 3 more lines of code and a bit hacky).

panekj:
Could you also add test for your issue similar to
https://github.com/tamasfe/taplo/blob/cb156c30c1072b40ed66cd1defe83975aa4ce1f1/crates/taplo/src/tests/formatter.rs#L161-L181

ia0 (approved):
I tried to play around to find when the `if` does not trigger but couldn't. So I agree we can move forward and wait until someone comes up with an example. In the absence of a test breaking, I think we can say that this is not an important feature that is being removed.

I don't think the `[` case is similar, because it can't come after a comment. The only thing that can follow a comment is a newline by definition of a comment. So only the occurrence in the NEWLINE case is actually problematic.

flying-sheep:
could we get a release with this?

ia0:
> could we get a release with this?

Ideally, I'd like #434 to get fixed before releasing. It's an important blocker for releasing to VS Code. Feel free to take a look at it if you're familiar with VS Code. PRs are welcome. Thanks!

flying-sheep:
I fixed that one, see #462

panekj (approved):

PR #465, opened 2023-08-23
Add Homebrew install to docs

uncenter:
There is a Homebrew formula (tamasfe#251) but it was never added to the docs.


ia0 (approved):
Thanks for the PR!

panekj (approved):

Issue #384, opened 2023-03-11
Poetry doesn't need author email.


real-yfprojects:
I have the following line in my `pyproject.toml`.

```toml
[tool.poetry]
authors = ["myname"]
```

This extension will show 
```
Pattern that matches Name <email> like 'King Arthur' or 'Miss Islington <miss-islington@python.org>'.
```
and
```
"myname" does not match "^(?:\S+?\s)+?(?:<(?:[a-z\d!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z\d!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z\d](?:[a-z\d-]*[a-z\d])?\.)+[a-z\d](?:[a-z\d-]*[a-z\d])?|\[(?:(?:25[0-5]|2[0-4][\d]|[01]?[\d][\d]?)\.){3}(?:25[0-5]|2[0-4][\d]|[01]?[\d][\d]?|[a-z\d-]*[a-z\d]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])>)?$"
```
although poetry doesn't require an author email since *v0.9.0*: python-poetry/poetry@5b14bd10e078dd8a8aa03732ee34ef9e65646021

ia0:
Fixed upstream.

real-yfprojects:
Thank you very much!

Issue #360, opened 2022-12-13
Extension complains about upper-case email addresses


Vuizur:
For example, the extension complains about the line in my pyproject.toml
authors = ["Vuizur <Vuizur@users.noreply.github.com>"]

But if I change it to:
authors = ["Vuizur <vuizur@users.noreply.github.com>"]
everything works again.

In my opinion an upper-case emails should not be a cause for complain (unless there are some details I am unaware of). 

PS: Thank you very much for developing this very useful extension!

ia0:
Fixed upstream.

Vuizur:
Awesome, thanks!

Dne st 13. 9. 2023 18:49 uživatel Julien Cretin ***@***.***>
napsal:

PR #467, opened 2023-09-08
Adapt regex from poetry project

SanderBeekhuis:
This PR solves #360  and  #384 

We adept the Regex in use in the poetry project  https://github.com/python-poetry/poetry-core/blob/main/src/poetry/core/packages/package.py in order to be just as permissive as the Poetry project, for which we are validating.

-----
This PR is our first contribution. We are very open to suggestions and improvements.

This PR was a group effort by Ordina Pythoneers, they would like to thank Ordina for the time to do this.

@SanderBeekhuis 
@RoelAdriaans
@Coen-Smulders

### DIFF
--- a/site/site/public/schemas/pyproject.toml.json
+++ b/site/site/public/schemas/pyproject.toml.json
@@ -9,7 +9,7 @@
     "poetry-author-pattern": {
       "description": "Pattern that matches `Name <email>` like 'King Arthur' or 'Miss Islington &lt;miss-islington@python.org&gt;'.",
       "type": "string",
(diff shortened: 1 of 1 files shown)


panekj:
I think it has to be fixed in SchemaStore since that's the primary source for schemas used
https://github.com/SchemaStore/schemastore/blob/94ae74640898dcdb162a68a963b6fb6a3a29ef9e/src/schemas/json/pyproject.json#L39

SanderBeekhuis:
@panekj (or somebody else).

It has been fixed upstream in https://github.com/SchemaStore/schemastore/pull/3209  what will  be the process of landing this fix in this extension?

ia0:
I don't think there's anything to do. The catalog and (if I'm not mistaken) the schemas are downloaded at runtime. So as soon as they publish the new version (which I think is quite fast and seems to already be the case now because the issues are fixed for me), then you should see the fix (possibly restarting your editor). I don't know if schemas are cached though. Please ping here if you it's not fixed.

SanderBeekhuis:
@ia0 You are right. It also works on my machine.

PR #471, opened 2023-09-13
Run clippy and fmt 

adrian-kong:
Some simple clippy cleanup, adds clippy and fmt to CI

These changes are entirely optional, just enforces some lints and reducing some boiler code

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -77,10 +77,10 @@ jobs:
           toolchain: stable
           override: true
       - uses: Swatinem/rust-cache@v1
-      - name: Clippy and format
-        run: |
-          cargo clippy --all-features --all-targets -- -D warnings
-          cargo fmt -- --check
+      - name: Clippy
+        run: cargo clippy --all-features --all-targets -- -D warnings
+      - name: Format
+        run: cargo fmt -- --check
       - name: Retrieve toml-test
         run: |
           wget https://github.com/BurntSushi/toml-test/releases/download/v1.1.0/toml-test-v1.1.0-linux-amd64.gz
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -77,6 +77,10 @@ jobs:
           toolchain: stable
           override: true
       - uses: Swatinem/rust-cache@v1
+      - name: Clippy and format
+        run: |
+          cargo clippy --all-features --all-targets -- -D warnings
+          cargo fmt -- --check
       - name: Retrieve toml-test
         run: |
           wget https://github.com/BurntSushi/toml-test/releases/download/v1.1.0/toml-test-v1.1.0-linux-amd64.gz
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ target
 .vscode/settings.json
 .cargo/config.toml
 yarn-error.log
+/.idea
--- a/crates/lsp-async-stub/src/lib.rs
+++ b/crates/lsp-async-stub/src/lib.rs
@@ -49,7 +49,7 @@ impl Cancellation {
     pub fn cancel(&mut self) {
         self.cancelled.store(true, Ordering::SeqCst);
 
-        if let Some(w) = std::mem::replace(&mut *self.waker.lock().unwrap(), None) {
+        if let Some(w) = (*self.waker.lock().unwrap()).take() {
             w.wake();
         }
     }
--- a/crates/lsp-async-stub/src/lib.rs
+++ b/crates/lsp-async-stub/src/lib.rs
@@ -269,8 +269,8 @@ impl<W: Clone> RequestWriter for Context<W> {
     }
 }
 
-pub trait MessageWriter: Sink<rpc::Message, Error = io::Error> + Unpin {}
-impl<T: Sink<rpc::Message, Error = io::Error> + Unpin> MessageWriter for T {}
+pub trait MessageWriter: Sink<rpc::Message, Error = io::Error> + Unpin + Send {}
+impl<T: Sink<rpc::Message, Error = io::Error> + Unpin + Send> MessageWriter for T {}
 
 struct Inner<W: Clone> {
     next_request_id: i32,
--- a/crates/lsp-async-stub/src/lib.rs
+++ b/crates/lsp-async-stub/src/lib.rs
@@ -14,6 +14,8 @@ use lsp_types::{
     NumberOrString,
 };
 use serde::{de::DeserializeOwned, Serialize};
+use std::cell::RefCell;
+use std::rc::Rc;
 use std::{
     collections::HashMap,
     io, mem,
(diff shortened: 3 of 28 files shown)


crates/lsp-async-stub/src/lib.rs
```diff
@@ -269,8 +269,8 @@ impl<W: Clone> RequestWriter for Context<W> {
     }
 }
 
-pub trait MessageWriter: Sink<rpc::Message, Error = io::Error> + Unpin {}
-impl<T: Sink<rpc::Message, Error = io::Error> + Unpin> MessageWriter for T {}
+pub trait MessageWriter: Sink<rpc::Message, Error = io::Error> + Unpin + Send {}
+impl<T: Sink<rpc::Message, Error = io::Error> + Unpin + Send> MessageWriter for T {}
```
adrian-kong, line 273: Makes `MessageWriter` require send.

crates/taplo-lsp/src/handlers/hover.rs
```diff
@@ -40,31 +40,24 @@ pub(crate) async fn hover<E: Environment>(
...
-            if p.syntax.kind() == IDENT || is_primitive(p.syntax.kind()) {
-                Some(p)
-            } else {
-                None
-            }
-        }) {
+        None => match query
+            .after
+            .clone()
+            .filter(|p| p.syntax.kind() == IDENT || is_primitive(p.syntax.kind()))
```
adrian-kong, line 59: Some changes unrelated to clippy, instead of `and_then` use `filter` here

.github/workflows/ci.yaml
```diff
@@ -77,6 +77,10 @@ jobs:
           toolchain: stable
           override: true
       - uses: Swatinem/rust-cache@v1
+      - name: Clippy and format
+        run: |
+          cargo clippy --all-features --all-targets -- -D warnings
+          cargo fmt -- --check
```
ia0, line 83: Does Github shell run with `-e`? Would this step fail if only clippy fails? Why not have 2 steps? One for clippy and one for format.

.gitignore
```diff
@@ -4,3 +4,4 @@ target
 .vscode/settings.json
 .cargo/config.toml
 yarn-error.log
+/.idea
```
ia0, line 7: I don't know this. Looks like some IDE thing. I'm a bit reticent to add such things. IDEs should be able to work without adding their stuff locally. That said, a few lines above I can see `.vscode` which is also an IDE which also seems to not work if it doesn't write stuff locally. So yeah... At least I like that you constrain the exception to the root of the repo.

crates/taplo-common/src/schema/associations.rs
```diff
@@ -256,9 +256,8 @@ impl<E: Environment> SchemaAssociations<E> {
 
     pub fn add_from_config(&self, config: &Config) {
         for rule in &config.rule {
-            let file_rule = match rule.file_rule.clone() {
-                Some(rule) => rule,
-                None => continue,
+            let Some(file_rule) = rule.file_rule.clone() else {
```
ia0, line 259: I'm a bit afraid this forces recent compiler version (and slightly a "breaking" change regarding MSRV). Does clippy really suggests to do this?

crates/taplo-wasm/src/lsp.rs
```diff
@@ -83,3 +83,5 @@ impl Sink<rpc::Message> for WasmLspInterface {
         std::task::Poll::Ready(Ok(()))
     }
 }
+
+unsafe impl Send for WasmLspInterface {}
```
ia0, line 87: Why is this safe? `js_sys::Function` doesn't seem to implement `Send`.

crates/taplo/src/dom/node/nodes.rs
```diff
@@ -19,7 +19,7 @@ macro_rules! wrap_node {
     ) => {
         $(#[$attrs])*
         $vis struct $name {
-            pub(crate) inner: Arc<$inner>,
+            pub(crate) inner: Rc<$inner>,
```
ia0, line 22: Why not use `Arc`?

ia0 (commented):
Thanks for the PR! Although I'm not sure what's the context and motivation for this. I left a few comments where it's not clear to me whether that should be done.

.gitignore
panekj: It's JetBrains IDEs, afaik unfortunately they add it all the time whenever you open any project and there is no way to not add it.

.github/workflows/ci.yaml
panekj: I would prefer clippy and rustfmt separate from each other, and clippy for each platform
Example: https://github.com/lapce/lapce/actions/runs/6130621751

crates/taplo/src/dom/node/nodes.rs
panekj: `Arc` is thread-safe, if threads are not a concern, `Rc` is better.

crates/taplo-common/src/schema/associations.rs
panekj: Seems to be https://rust-lang.github.io/rust-clippy/stable/index.html#option_if_let_else

crates/taplo-wasm/src/lsp.rs
adrian-kong: Oh actually this isn't safe, I made a mistake thinking the writer also required to implement Send which is not the case. The `Sink` MessageWriter type doesn't require Send so it might be better not to have `Arc<Mutex<>>` lock instead replace with more performant Rc and fix clippy suggestions?

crates/taplo/src/dom/node/nodes.rs
adrian-kong: Yeah, was wondering what the intention here was since I'm new to the codebase. It looked like `Arc` implementation can be replaced with `Rc` here as we don't enforce `Send` in the inner types. If these are breaking changes, I can revert these type changes as well as the `let else` changes which were stabilised in 1.65 rust.

crates/taplo-common/src/schema/associations.rs
adrian-kong: Hmm yeah, this let-else was stabilised in 1.65 rust which would be breaking for old versions.

.gitignore
adrian-kong: Yeah, the JetBrains `/.idea` folder is entirely unnecessary [for versions beyond 2019.1](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839)

I could remove this change if necessary, usually `/.idea` folder doesn't get tracked as it can be autogenerated
https://stackoverflow.com/questions/43198273/which-files-in-idea-folder-should-be-tracked-by-git

crates/taplo/src/dom/node/nodes.rs
ia0: I would keep `Arc` to preserve usage in multi-threaded environment (there is some in the LSP and CLI as far as I remember). The benefit of `Rc` is useless because Taplo does not need performance. The project needs to be easily maintainable, safe, and correct before all.

I would also not use `let else` just yet. It's also not a huge benefit (actually I'm not even convinced it improves readability at all).

From a general point of view, Taplo is only in maintenance mode, so only critical changes are justified. However, if you have time to invest in the project, then I think it would make sense to make you a maintainer and you would then have more freedom to address larger code changes.

adrian-kong: Sounds good, got sidetracked but originally came for the issue which was resolved in #474

adrian-kong:
Closing as non critical change

PR #489, opened 2023-10-11
Add option to specify separator for text output

mkatychev:
Added `--separator <str>` flag for `taplo get`.

From the helpdoc:
```
 --separator <SEPARATOR>
        A string that separates array values when printing to stdout.

        If `--separator` is specified with a non text output format, the operation will fail.
```

### DIFF
--- a/crates/taplo-cli/src/commands/queries.rs
+++ b/crates/taplo-cli/src/commands/queries.rs
@@ -103,9 +103,8 @@ impl<E: Environment> Taplo<E> {
                 }
             }
             crate::args::OutputFormat::Value => {
-                let mut buf = String::new();
                 let separator = cmd.separator.as_deref().unwrap_or("\n");
-                if let Some(p) = cmd.pattern {
+                let mut buf = if let Some(p) = cmd.pattern {
                     let p = p.trim_start_matches('.');
 
                     let nodes = p
@@ -118,13 +117,13 @@ impl<E: Environment> Taplo<E> {
                     }
 
                     let values = nodes
-                        .map(|(_, node)| extract_value(&node, &separator))
+                        .map(|(_, node)| extract_value(&node, separator))
                         .collect::<Result<Vec<String>, _>>()?;
 
-                    buf += &values.join(&separator);
+                    values.join(&separator)
                 } else {
-                    buf = extract_value(&node, &separator)?;
-                }
+                    extract_value(&node, separator)?
+                };
 
                 if !cmd.strip_newline {
                     buf += "\n";
--- a/crates/taplo-cli/src/commands/queries.rs
+++ b/crates/taplo-cli/src/commands/queries.rs
@@ -104,7 +104,7 @@ impl<E: Environment> Taplo<E> {
             }
             crate::args::OutputFormat::Value => {
                 let mut buf = String::new();
-                let separator = cmd.separator.unwrap_or("\n".to_string());
+                let separator = cmd.separator.as_deref().unwrap_or("\n");
                 if let Some(p) = cmd.pattern {
                     let p = p.trim_start_matches('.');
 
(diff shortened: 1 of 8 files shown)


crates/taplo-cli/src/commands/queries.rs
```diff
@@ -14,6 +17,11 @@ impl<E: Environment> Taplo<E> {
     pub async fn execute_get(&self, cmd: GetCommand) -> Result<(), anyhow::Error> {
         let mut stdout = self.env.stdout();
 
+        // `--separator` should only be handled for a text output for mat
+        if cmd.separator.is_some() && !matches!(cmd.output_format, OutputFormat::Value) {
```
mkatychev, line 21: I figured making `cmd.separator` an `Option<String>` would catch instances where the end user provided a separator for non text formats.

I wasn't able to find a "requires other arg to be this value" in clap to guard against this:
https://docs.rs/clap/latest/clap/struct.Arg.html

crates/taplo-cli/src/commands/queries.rs
```diff
@@ -93,52 +103,35 @@ impl<E: Environment> Taplo<E> {
                 }
             }
             crate::args::OutputFormat::Value => {
+                let mut buf = String::new();
+                let separator = cmd.separator.unwrap_or("\n".to_string());
```
ia0, line 107: It doesn't look like this needs to be owned. What about borrowing only?

```suggestion
                let separator = cmd.separator.as_deref().unwrap_or("\n");
```

crates/taplo-cli/src/commands/queries.rs
```diff
@@ -93,52 +103,35 @@ impl<E: Environment> Taplo<E> {
...
-
-                    if cmd.strip_newline {
-                        if buf.ends_with('\n') {
-                            let new_len = buf.trim_end().len();
-                            buf.truncate(new_len);
-                        }
-                    } else if !buf.ends_with('\n') {
-                        buf += "\n";
-                    }
+                    buf = extract_value(&node, &separator)?;
```
ia0, line 126: This seems to discard anything done before. Shouldn't this instead be the following?

```suggestion
                    buf += &extract_value(&node, &separator)?;
```

ia0 (changes requested):
Thanks for the PR!

Out of curiosity, what's the motivation behind this change? Which separator do you want to use and in which context?

crates/taplo-cli/src/commands/queries.rs
mkatychev: It seems strange to append when the else statement is guaranteed to call `extract_value` once 🤔 , I'll look into making `buf` be non mutable.

crates/taplo-cli/src/commands/queries.rs
```diff
@@ -93,52 +103,35 @@ impl<E: Environment> Taplo<E> {
...
-                    if cmd.strip_newline {
-                        if buf.ends_with('\n') {
-                            let new_len = buf.trim_end().len();
-                            buf.truncate(new_len);
-                        }
-                    } else if !buf.ends_with('\n') {
-                        buf += "\n";
-                    }
+                    let values = nodes
+                        .map(|(_, node)| extract_value(&node, &separator))
```
ia0, line 121: You can now replace all occurrences of `&separator` by `separator`.

crates/taplo-cli/src/commands/queries.rs
ia0: My bad, I thought you were moving the `buf` out of a loop, but it's just out of an `if`. I guess there's a few options. Here's another one:

```rust
let mut buf;
if let Some(...) = ... {
    buf = String::new();
    buf += ...
} else {
    buf = extract_value(...);
}
...
```

mkatychev:
> Thanks for the PR!
> 
> Out of curiosity, what's the motivation behind this change? Which separator do you want to use and in which context?

This would be to pass in the result of `taplo get -f rust-toolchain.toml 'toolchain.components'` to the rust toolchain github action, the GHA asks for a comma separated string to installs said components

ia0 (approved):
Thanks for the PR and the motivation explanation!

PR #494, opened 2023-10-19
chore: make link in settings description clickable

uncenter:
For some reason you have to make links in settings descriptions with markdown syntax for them to turn into clickable links.

Screenshot 2023-10-18 at 21 05 28 (Visual Studio Code)

*The link in the bottom right, which uses the bracket parenthesis syntax (`[123](https://example.com)`), is underlined and clickable while the plain link at the top center is not converted into a link and has to copy pasted.*

I just took the title of the docs page and the section heading as the link text.

### DIFF
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -387,7 +387,7 @@
           "scope": "resource",
           "type": "array",
           "default": [],
-          "description": "Array of Taplo rules in JSON format, see [Configuration File - Rules](https://taplo.tamasfe.dev/configuration/file.html#rules). The rules given here are appended to existing rules from config files, if any. There is no conversion done, all object keys must be snake_case, including formatting rules."
+          "markdownDescription": "Array of Taplo rules in JSON format, see [Configuration File - Rules](https://taplo.tamasfe.dev/configuration/file.html#rules). The rules given here are appended to existing rules from config files, if any. There is no conversion done, all object keys must be snake_case, including formatting rules."
         }
       }
     }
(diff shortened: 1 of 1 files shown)


editors/vscode/package.json
```diff
@@ -387,7 +387,7 @@
           "scope": "resource",
           "type": "array",
           "default": [],
-          "description": "Array of Taplo rules in JSON format, see https://taplo.tamasfe.dev/configuration/file.html#rules. The rules given here are appended to existing rules from config files, if any. There is no conversion done, all object keys must be snake_case, including formatting rules."
+          "description": "Array of Taplo rules in JSON format, see [Configuration File - Rules](https://taplo.tamasfe.dev/configuration/file.html#rules). The rules given here are appended to existing rules from config files, if any. There is no conversion done, all object keys must be snake_case, including formatting rules."
```
ia0, line 390: You can also wrap a URL between `<` and `>` to make it clickable.

ia0 (approved):
Thanks for the PR!

editors/vscode/package.json
uncenter: I considered that but the link is really long so I thought a title was best. Any preference?

ia0: I don't know VSCode. If the link is always clickable then keeping it like that is fine and indeed more readable. However if it's possible that for some reason it's not clickable, then it's better to use `<>` for users to copy/paste it.

uncenter: Yeah it should be clickable with the `[]()` syntax.

panekj (approved):

uncenter:
Wait let me just verify this works real quick.

panekj (approved):

Issue #473, opened 2023-09-15
Release a new version


ia0:
#449 and #462 are fixing important issues and need to be released. There are other minor changes since the last release.

Before releasing, we should also bump `json_value_merge` to 2.0.0 (and update the code accordingly) because 1.2.0 was a breaking change.

ia0:
I created #474 for the `json_value_merge` issue.

panekj:
> It would be great because [CI fails](https://github.com/lifescapegame/lifescape/actions/runs/6147934293/job/16680499099?pr=191) for me when I trying to install Taplo.

You should always use `--locked` flag when installing with `cargo install`

Shatur:
I would like to always install the latest version.

ia0:
The problem here is that you can't install with `--locked` in nightly because you need #449 (`proc-macro` issue) and you can't install without `--locked` because you need #474 (`json_value_merge` issue). So right now if you want to install Taplo in CI you need to install with stable and `--locked` or from git if you use nightly. This is a considerable issue we are aware of and a release is needed for many other reasons.

@tamasfe Do you think you would have time for a release? On my side, I'm not sure I do. @panekj what about you?

David-Else:
I hope the full language server version can be released via NPM too rather than the rather useless version that is currently there. Cheers :)

panekj:
> what about you?

I'll take a look at the process tonight but nothing promised.

flying-sheep:
Are you making any headway with this? There’s now multiple bugfixes to CLI and extension that I’m waiting for.

ia0:
Would be nice to have #496 merged before the release, such that we also produce builds for ARM.

flying-sheep:
Eventually, you might want to switch to https://release-plz.ieni.dev/ which automates a lot of things.

JounQin:
The npm package is also very outdated #480

o-az:
If anyone's interested in forking this and maintaining it with me let me know.

JounQin:
I've forked it at https://github.com/unrs/taplo but isn't not sure to publish it.

ia0:
Feel free to also volunteer at [maintaining this crate](https://github.com/tamasfe/taplo/issues/403). That's probably the most natural way to start contributing. You would be able to fork later if something is limiting you.

If you nevertheless want to fork, which I would understand if you are planning to invest significant effort refactoring, I would suggest contacting @tamasfe to discuss the best way to proceed:
[...]

PR #496, opened 2023-10-26
ci: build docker images for linux/arm64

nikaro:
* Add `linux/arm64` in the build platforms
* Use a matrix for parallel build of Docker images
* Update parent images to `rust:1.73-alpine3.18 as build` and `alpine:3.18`

### DIFF
--- a/.github/workflows/releases.yaml
+++ b/.github/workflows/releases.yaml
@@ -261,6 +261,13 @@ jobs:
     runs-on: ubuntu-latest
     needs: ["wait_for_ci"]
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/release-taplo-cli-0')
+    strategy:
+      matrix:
+        image:
+          - file: docker/cli-alpine.Dockerfile
+            tags: tamasfe/taplo:latest,tamasfe/taplo:${{ env.RELEASE_VERSION }},tamasfe/taplo:${{ env.RELEASE_VERSION }}-alpine
+          - file: docker/cli-full-alpine.Dockerfile
+            tags: tamasfe/taplo:full,tamasfe/taplo:${{ env.RELEASE_VERSION }}-full,tamasfe/taplo:${{ env.RELEASE_VERSION }}-full-alpine
     steps:
       - uses: actions/checkout@v3
       - name: Retrieve release version
@@ -275,22 +282,14 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-      - name: Build and push CLI alpine
+      - name: Build and push
         uses: docker/build-push-action@v3
         with:
-          file: "docker/cli-alpine.Dockerfile"
-          context: "."
-          platforms: linux/amd64
+          file: ${{ matrix.image.file }}
+          context: .
+          platforms: linux/amd64,linux/arm64
           push: true
-          tags: tamasfe/taplo:latest,tamasfe/taplo:${{ env.RELEASE_VERSION }},tamasfe/taplo:${{ env.RELEASE_VERSION }}-alpine
-      - name: Build and push CLI Full alpine
-        uses: docker/build-push-action@v3
-        with:
-          file: "docker/cli-full-alpine.Dockerfile"
-          context: "."
-          platforms: linux/amd64
-          push: true
-          tags: tamasfe/taplo:full,tamasfe/taplo:${{ env.RELEASE_VERSION }}-full,tamasfe/taplo:${{ env.RELEASE_VERSION }}-full-alpine
+          tags: ${{ matrix.image.tags }}
 
   build_cli_windows:
     name: ${{ matrix.triple }}
--- a/docker/cli-alpine.Dockerfile
+++ b/docker/cli-alpine.Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.61-alpine3.16 as build
+FROM rust:1.73-alpine3.18 as build
 
 WORKDIR /build
 
@@ -8,7 +8,7 @@ COPY . .
 
 RUN cargo build -p taplo-cli --release
 
-FROM alpine:3.16
+FROM alpine:3.18
 
 COPY --from=build /build/target/release/taplo /usr/bin/taplo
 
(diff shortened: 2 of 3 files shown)


ia0 (approved):
Thanks for the PR! This technically looks good to me, however I would like to know first what motivates this change.

In particular, there's 2 aspects I'm uncertain about:
1. How costly are the steps that are now duplicated because of the matrix. What do we gain by using a matrix?
2. Does Taplo guarantee MSRV? It doesn't look like it but I might be wrong. If it does, then maybe 1.61 was chosen for a reason.

panekj:
For point 1, using matrix makes jobs run in parallel instead on a single runner one after another.

panekj:
> > For point 1, using matrix makes jobs run in parallel instead on a single runner one after another.
> 
> Yes but that's not the question. The question is whether the 3 docker actions that will now be duplicated are costly or not. (And implicitly if we can afford that cost.)

I'm not sure what cost are you speaking of

nikaro:
> Thanks for the PR! This technically looks good to me, however I would like to know first what motivates this change.

This is mainly to be able to use Taplo through Docker on ARM machines, like the Apple Silicon MacBooks or Raspberry Pi.
I want to make a pre-commit hook that uses the docker image to lint and format TOML file, but currently it wont work on M1 machine.

> In particular, there's 2 aspects I'm uncertain about:
> 
> 1. How costly are the steps that are now duplicated because of the matrix. What do we gain by using a matrix?

Most of the time for this job is consumed by the `docker/build-push-action@v3` step, there others only take a few seconds each. So it won't change much regarding CI minutes consumption.

> 2. Does Taplo guarantee MSRV? It doesn't look like it but I might be wrong. If it does, then maybe 1.61 was chosen for a reason.

I don't speak Rust 😬 so i don't understand what you are talking about. But i upgraded the `FROM rust:*` image because i couldn't build the container with the current version:

```
> docker build -t taplo -f docker/cli-alpine.Dockerfile .
[+] Building 267.6s (11/12)                                                                                                                                                                 docker:orbstack
 => [internal] load build definition from cli-alpine.Dockerfile                                                                                                                                        0.0s
 => => transferring dockerfile: 297B                                                                                                                                                                   0.0s
 => [internal] load .dockerignore                                                                                                                                                                      0.0s
 => => transferring context: 47B                                                                                                                                                                       0.0s
 => [internal] load metadata for docker.io/library/alpine:3.16                                                                                                                                         1.5s
 => [internal] load metadata for docker.io/library/rust:1.61-alpine3.16                                                                                                                                1.0s
 => [build 1/5] FROM docker.io/library/rust:1.61-alpine3.16@sha256:3bcdeab61ea4e01db688277c5565b4bfb845293520d3b5eee7e9acf02ad6902b                                                                    0.0s
 => CACHED [stage-1 1/2] FROM docker.io/library/alpine:3.16@sha256:a8cbb8c69ee71561f4b69c066bad07f7e510caaa523da26fbfc606b10bd7934b                                                                    0.0s
 => [internal] load build context                                                                                                                                                                      0.0s
 => => transferring context: 91.75kB                                                                                                                                                                   0.0s
 => CACHED [build 2/5] WORKDIR /build                                                                                                                                                                  0.0s
 => CACHED [build 3/5] RUN apk add --no-cache musl-dev                                                                                                                                                 0.0s
 => [build 4/5] COPY . .                                                                                                                                                                               0.1s
 => ERROR [build 5/5] RUN cargo build -p taplo-cli --release                                                                                                                                         266.0s
------
 > [build 5/5] RUN cargo build -p taplo-cli --release:
[...]
251.3    Compiling reqwest v0.11.12
252.7 error[E0658]: use of unstable library feature 'bool_to_option'
252.7    --> crates/taplo/src/formatter/mod.rs:608:18
252.7     |
252.7 608 |                 .then_some(*indent + 1)
252.7     |                  ^^^^^^^^^
252.7     |
252.7     = note: see issue #80967 <https://github.com/rust-lang/rust/issues/80967> for more information
252.7
253.8 For more information about this error, try `rustc --explain E0658`.
253.8 error: could not compile `taplo` due to previous error
253.8 warning: build failed, waiting for other jobs to finish...
------
cli-alpine.Dockerfile:9
--------------------
   7 |     COPY . .
   8 |
   9 | >>> RUN cargo build -p taplo-cli --release
  10 |
  11 |     FROM alpine:3.16
--------------------
ERROR: failed to solve: process "/bin/sh -c cargo build -p taplo-cli --release" did not complete successfully: exit code: 101
```

panekj:
This has been stabilized in 1.62
```
252.7 error[E0658]: use of unstable library feature 'bool_to_option'
252.7    --> crates/taplo/src/formatter/mod.rs:608:18
252.7     |
252.7 608 |                 .then_some(*indent + 1)
252.7     |                  ^^^^^^^^^
252.7     |
252.7     = note: see issue #80967 <https://github.com/rust-lang/rust/issues/80967> for more information
```

ia0:
> This is mainly to be able to use Taplo through Docker on ARM machines

I see, so that explains the added linux/arm64 line.

> Most of the time for this job is consumed by the docker/build-push-action@v3 step, the others steps only take a few seconds each.

Ok perfect, so the factorization is fine.

> I don't speak Rust

This is the Minimum Supported Rust Version. Some projects like to guarantee that their project builds for old versions of the compiler (as long as it's not older than the MSRV they want to support). I couldn't find any mention of the MSRV for Taplo (which kind of makes sense because it's much for of a binary than a library, and MSRV doesn't make too much sense for a binary).

So all good on my side to merge. That said, I couldn't find a successful run of this job, so I'm not sure this PR will solve your problem, but at least it increases the chances of solving your problem. Thanks again for the PR!

nikaro:
> So all good on my side to merge. That said, I couldn't find a successful run of this job, so I'm not sure this PR will solve your problem, but at least it increases the chances of solving your problem. Thanks again for the PR!

Here is the last run for this job: https://github.com/tamasfe/taplo/actions/runs/5511385857/job/14920644021c
The log has expired but its outcome is successful so it should be fine. But we won't know it before the next `release-taplo-cli-0*` tag push.

Thanks for the quick feedback! 🙏

ia0:
> Here is the last run for this job: https://github.com/tamasfe/taplo/actions/runs/5511385857/job/14920644021c 

Cool, thanks! I missed it. So it's 19min and was the last release so yeah it should probably work for the next as well, which should ideally be soon (see #473).

panekj (approved):

PR #498, opened 2023-10-30
Use JSON Schema draft 2020-12

### LINKED ISSUE #497: Discrepancy in validation with JSON Schema draft 2020-12
With this schema, which defines an object with a property `test` that contains an array of objects that can have certain properties, I am having issues where Taplo should be invalidating a test case but it isn't.

```json
{
	"$schema": "https://json-schema.org/draft/2020-12/schema",
	"type": "object",
	"properties": {
		"test": {
			"type": "array",
			"items": {
				"$comment": "This schema is closed, and any property used which aren't defined are invalid/not allowed.",
				"type": "object",
				"properties": {
					"foo": { "type": "string" }
				},
				"allOf": [{ "$ref": "#/$defs/bazExtension" }],
				"unevaluatedProperties": false
			}
		}
	},
	"$defs": {
		"bazExtension": {
			"$comment": "This schema can be used as an extension",
			"type": "object",
			"properties": {
				"baz": { "type": "boolean" }
			}
		}
	}
}
```

These tests should be valid, any combination of `foo` and `baz` together or separate:

```json
{
	"test": [
		{
			"foo": "foo",
			"baz": true
		}
	]
}
```

```json
{
	"test": [
		{
			"foo": "foo"
		}
	]
}
```

```json
{
	"test": [
		{
			"baz": true
		}
	]
}
```

```toml
[[test]]
foo = "foo"
baz = true
```

```toml
[[test]]
foo = "foo"
```

```toml
[[test]]
baz = true
```

And these should be invalid, with an additional property not in the definition:

```json
{
	"test": [
		{
			"foo": "foo",
			"baz": true,
			"otherProperty": true
		}
	]
}
```

```toml
[[test]]
foo = "foo"
baz = true
otherProperty = true
```

You can test these on https://json-everything.net/json-schema/ for JSON and with `taplo` obviously for TOML. My issue is that Taplo isn't invalidating the last example with an additional property, `otherProperty`, as it should be doing.

ia0:
I'm not sure we want to merge this. Not sure of the implications and why `jsonschema` doesn't figure out the draft version based on the `$schema` value in the schema.

Fixes #497

### DIFF
--- a/crates/taplo-common/Cargo.toml
+++ b/crates/taplo-common/Cargo.toml
@@ -21,7 +21,7 @@ hex = "0.4.3"
 indexmap = { version = "1.6.0", features = ["serde", "rayon"] }
 itertools = "0.10.3"
 json_value_merge = "2.0.0"
-jsonschema = { version = "0.17.1", default-features = false }
+jsonschema = { version = "0.17.1", default-features = false, features = ["draft202012"] }
 lru = "0.7.2"
 parking_lot = "0.12.0"
 percent-encoding = "2.1.0"
--- a/crates/taplo-common/src/schema/mod.rs
+++ b/crates/taplo-common/src/schema/mod.rs
@@ -260,6 +260,7 @@ impl<E: Environment> Schemas<E> {
             .with_resolver(CacheSchemaResolver {
                 cache: self.cache().clone(),
             })
+            .with_draft(jsonschema::Draft::Draft202012)
             .with_format("semver", formats::semver)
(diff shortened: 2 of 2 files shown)


salim-b:
I think this PR became obsolete with jsonschema-rs 0.19.0, cf. https://github.com/tamasfe/taplo/issues/497#issuecomment-2351065113.

ia0:
Thanks! I'm closing it.

Issue #506, opened 2023-11-29
feature request: support loading additional configurations for LSP


cdmistman:
I'm a Developer Experience Engineer at https://replit.com and we'd like to use taplo for configuring LSP support for our `.replit` file

Right now the fact that `taplo lsp` currently only respects 1 configuration file between `$PWD` and `/` is a blocker. Ideally, we'd be able to start the lsp like:
```sh
taplo lsp --load-config /path/to/.replit/schema/config.toml stdio
```

This would allow us to specify the schema to use for all `.replit` paths as well as enabling the user to extend the LSP with additional schemas. Additionally, it would allow us to use Nix to configure the LSP (our preferred configuration method) as the `config.toml` file (and schema) would be in `/nix/store`.

ia0:
Thanks for opening an issue! If I understand correctly, you want the `--config` flag of `taplo format` to also be available in `taplo lsp`. If that's correct, I'm not sure I see any issue with that. Probably the whole `GeneralArgs` struct could be flattened in `taplo lsp`. However, given this project is in maintenance mode, new features won't be worked on. Feel free to open a PR and we'll review it.

cdmistman:
i went ahead and added all of GeneralArgs to `taplo lsp` - lmk in the [PR](#510) if there's anything needed to get it merged and for another release to be cut! thank you!

cdmistman:
note: would still be nice to support multiple config files, but that PR is the only blocker for our purposes for now :)

PR #510, opened 2023-12-07
add GeneralArgs support to lsp command

cdmistman:
see #506 - this only supports 1 config file being passed in (the same behavior as other taplo commands) but it should do the trick for our desired case :)


ia0 (approved):
Thanks!

cdmistman:
@JounQin you seem to be the most active maintainer now - any chance in merging this before the next taplo release now that releases have been figured out?

JounQin:
Thanks for your contribution!

Issue #511, opened 2023-12-08
Inconsistent alphabetical ordering of table entries


IttayWeiss:
The toml snippet: 
'''
    a = { t = true }
    a- = { t = true }

    b-.t = true
    b.t = true
'''
shows the inconsistency. Semantically, there is no difference between s = { t = true } and s.t = true, so the two groups should be ordered in the same way. The alphabetical ordering should be by field name, and not by the string preceding the '='.

ia0:
Thanks for the bug report! I can reproduce.

The project is passively maintained which means this will only get fixed if someone writes a PR. The sorting code needs some cleanup by using a helper function do transform the key for sorting. That cleanup function could suffix a dot. That should fix this issue.

PR #513, opened 2023-12-11
Sort by crate lexicographical order

bebert64:
Fixes [#511](https://github.com/tamasfe/taplo/issues/511).
-----
First time contributing to an open-source project, so please let me know if I've done anything wrong or how I can improve. Also should I squash my commits before submitting the PR?
And running cargo fmt completely changes the way the imports are structured, so I ended up saving without formatting, but I'm not sure if that's what's intended

### DIFF
--- a/crates/taplo/src/formatter/mod.rs
+++ b/crates/taplo/src/formatter/mod.rs
@@ -395,7 +395,9 @@ use formatted_entry::FormattedEntry;
 
 impl PartialEq for FormattedEntry {
     fn eq(&self, other: &Self) -> bool {
-        self.cleaned_key().eq(other.cleaned_key())
+        self.cleaned_key()
+            .split('.')
+            .eq(other.cleaned_key().split('.'))
     }
 }
 
--- a/crates/taplo/src/formatter/mod.rs
+++ b/crates/taplo/src/formatter/mod.rs
@@ -9,9 +9,9 @@ use {
         syntax::{SyntaxElement, SyntaxKind::*, SyntaxNode, SyntaxToken},
         util::overlaps,
     },
+    once_cell::unsync::OnceCell,
     rowan::{GreenNode, NodeOrToken, TextRange},
     std::{
-        cell::RefCell,
         cmp,
         iter::{repeat, FromIterator},
         ops::Range,
@@ -361,57 +361,40 @@ fn format_impl(node: SyntaxNode, options: Options, context: Context) -> String {
 struct FormattedEntry {
     syntax: SyntaxElement,
     key: String,
-    _cleaned_key: RefCell<Option<String>>,
+    _cleaned_key: OnceCell<String>,
     value: String,
     comment: Option<String>,
 }
 
 impl PartialEq for FormattedEntry {
     fn eq(&self, other: &Self) -> bool {
-        self.cleaned_key().eq(&other.cleaned_key())
+        self.cleaned_key().eq(other.cleaned_key())
     }
 }
 
 impl Eq for FormattedEntry {}
 
 impl FormattedEntry {
-    fn cleaned_key<'a>(&'a self) -> std::cell::Ref<'a, Option<String>> {
-        self._cleaned_key
-            .borrow_mut()
-            .get_or_insert_with(|| self.key.replace('\'', "").replace('"', ""));
-        self._cleaned_key.borrow()
+    fn cleaned_key(&self) -> &str {
+        &self
+            ._cleaned_key
+            .get_or_init(|| self.key.replace('\'', "").replace('"', ""))
     }
 }
 
 impl PartialOrd for FormattedEntry {
     fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
         self.cleaned_key()
-            .as_deref()
-            .expect("Has been initialized")
             .split('.')
-            .partial_cmp(
-                &mut other
-                    .cleaned_key()
-                    .as_deref()
-                    .expect("Has been initialized")
-                    .split('.'),
-            )
+            .partial_cmp(&mut other.cleaned_key().split('.'))
     }
 }
 
 impl Ord for FormattedEntry {
     fn cmp(&self, other: &Self) -> cmp::Ordering {
         self.cleaned_key()
-            .as_deref()
-            .expect("Has been initialized")
             .split('.')
-            .cmp(
-                &mut other
-                    .cleaned_key()
-                    .as_deref()
-                    .expect("Has been initialized")
-                    .split('.'),
-            )
+            .cmp(&mut other.cleaned_key().split('.'))
     }
 }
 
@@ -788,7 +771,7 @@ fn format_entry(node: SyntaxNode, options: &Options, context: &Context) -> Forma
 
     FormattedEntry {
         syntax: node.into(),
-        _cleaned_key: RefCell::new(None),
+        _cleaned_key: OnceCell::new(),
         key,
         value,
         comment,
--- a/crates/taplo/src/formatter/mod.rs
+++ b/crates/taplo/src/formatter/mod.rs
@@ -375,39 +375,22 @@ impl PartialEq for FormattedEntry {
 
 impl Eq for FormattedEntry {}
 
+fn cleaned_key(key: &str) -> String {
+    key.replace('\'', "").replace('"', "")
+}
+fn split_key<'a>(key: &'a str) -> std::str::Split<'a, char> {
+    key.split('.').into_iter()
+}
+
 impl PartialOrd for FormattedEntry {
     fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
-        self.key
-            .replace('\'', "")
-            .replace('"', "")
-            .split('.')
-            .into_iter()
-            .partial_cmp(
-                &mut other
-                    .key
-                    .replace('\'', "")
-                    .replace('"', "")
-                    .split('.')
-                    .into_iter(),
-            )
+        split_key(&cleaned_key(&self.key)).partial_cmp(&mut split_key(&cleaned_key(&other.key)))
     }
 }
 
 impl Ord for FormattedEntry {
     fn cmp(&self, other: &Self) -> cmp::Ordering {
-        self.key
-            .replace('\'', "")
-            .replace('"', "")
-            .split('.')
-            .into_iter()
-            .cmp(
-                &mut other
-                    .key
-                    .replace('\'', "")
-                    .replace('"', "")
-                    .split('.')
-                    .into_iter(),
-            )
+        split_key(&cleaned_key(&self.key)).cmp(&mut split_key(&cleaned_key(&other.key)))
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/taplo/src/formatter/mod.rs
```diff
@@ -358,36 +359,40 @@ fn format_impl(node: SyntaxNode, options: Options, context: Context) -> String {
 struct FormattedEntry {
     syntax: SyntaxElement,
     key: String,
+    _cleaned_key: OnceCell<String>,
```
ia0, line 362: In Rust, underscores are usually used to indicate that something is not used. I know some other languages use it to indicate something is private. But Rust already has a mechanism for that. So, let's remove the underscore. It's not possible to confuse it with the function of the same name because they have different types, if that was the concern.

ia0 (changes requested):
Thanks for the PR!

For coherence, I think we should use the same value for Eq and Ord (even though for well-formed keys it shouldn't matter for the current PR):
- either the cleaned key (`String`)
- or the dot-separated cleaned key (`Vec<String>`)

crates/taplo/src/formatter/mod.rs
bebert64: You're right. I haven't been coding for a long time, and where I do, we kinda have this convention that it can be used as either "not used" or "should not be used directly", which really is what I was going for. I realize the types make it so no confusion is possible, but it still feels like we should indicate in some way that one should not use the value directly and/or try to use its own `get_or_init` (which could potentially make for hard-to-debug issues later on). So in the end, I have now made the var entirely private :smile: . Let me know what you think.

crates/taplo/src/formatter/mod.rs
```diff
@@ -355,39 +358,62 @@ fn format_impl(node: SyntaxNode, options: Options, context: Context) -> String {
     formatted
 }
 
-struct FormattedEntry {
-    syntax: SyntaxElement,
-    key: String,
-    value: String,
-    comment: Option<String>,
+mod formatted_entry {
```
panekj, line 361: Is there a reason for it to be separate module?

bebert64: The idea is to keep cleaned_key private, to make sure no one can initialize it any other way than using `replace().replace()` that was previously defined. Unless of course, you find it a bit overkill?

crates/taplo/src/formatter/mod.rs
ia0: I think I preferred the previous version :joy: 

I don't think the additional readability cost is justified. At best, I would just add a comment on top of the field, stating it's caching the key in a format suitable for comparison and should be accessed through the helper method.

bebert64:
> Thanks for the PR!
> 
> For coherence, I think we should use the same value for Eq and Ord (even though for well-formed keys it shouldn't matter for the current PR):
> 
>     * either the cleaned key (`String`)
> 
>     * or the dot-separated cleaned key (`Vec<String>`)

The whole issue being that we need to compare the Split instead of the String itself, I believe the only option is then to use the Split for Eq as well.

crates/taplo/src/formatter/mod.rs
panekj: You can keep all other fields public without making `cleaned_key` public, no?

ia0:
> The whole issue being that we need to compare the Split instead of the String itself, I believe the only option is then to use the Split for Eq as well.

Yes, we should use the same value for all comparisons: Eq and Ord. The choice between `"foo.bar"` and `["foo", "bar"]` doesn't make a big difference to me. It's true the second is a bit cleaner, but the first is also simpler. I'm fine with both.

bebert64:
> > The whole issue being that we need to compare the Split instead of the String itself, I believe the only option is then to use the Split for Eq as well.
> 
> Yes, we should use the same value for all comparisons: Eq and Ord. The choice between `"foo.bar"` and `["foo", "bar"]` doesn't make a big difference to me. It's true the second is a bit cleaner, but the first is also simpler. I'm fine with both.

I'm not sure I understand. If we use "foo.bar" for Ord, don't we have exactly the same problem as before?

crates/taplo/src/formatter/mod.rs
bebert64: I'm not sure how to do that without the extra module layer. As it is right, now, it's already 100% private to this module (`formatter::mod.rs`). @ia0 seems to prefer a simple comment anyway, so I'm currently preparing this simplified version. Let me know what you think as well

ia0:
> I'm not sure I understand. If we use "foo.bar" for Ord, don't we have exactly the same problem as before?

Sorry, what I meant was `"foo.bar."` like I initially suggested in https://github.com/tamasfe/taplo/issues/511#issuecomment-1846912854.

bebert64:
> Sorry, what I meant was `"foo.bar."` like I initially suggested in [#511 (comment)](https://github.com/tamasfe/taplo/issues/511#issuecomment-1846912854).

I apologize, I still must have misunderstood something. From what I had gathered, the problem comes from the fact that the `'-'` comes before the `'.'`, which leads to `"foo-.bar"` to come before `"foo.bar"`, even though `"foo"` comes before `"foo-"` (and we wanted those two ordering to be consistent). If that is correct, the `"foo.bar."` will still come after `"foo-.bar."`, no? On the other hand, `["foo", "bar"]` would correctly come before `["foo-", "bar"]`.

bebert64:
Ok, got it now, thanks. Then I think I don't have new modifications to write. I'm not sure what's the correct process now: should I re-ask for a review or is that implied by this conversation?

crates/taplo/src/formatter/mod.rs
```diff
@@ -358,36 +359,42 @@ fn format_impl(node: SyntaxNode, options: Options, context: Context) -> String {
...
+
 impl PartialEq for FormattedEntry {
     fn eq(&self, other: &Self) -> bool {
-        self.key
-            .replace('\'', "")
-            .replace('"', "")
-            .eq(&other.key.replace('\'', "").replace('"', ""))
+        self.cleaned_key()
+            .split('.')
+            .eq(other.cleaned_key().split('.'))
```
ia0, line 381: No need to split.

crates/taplo/src/formatter/mod.rs
```diff
@@ -358,36 +359,42 @@ fn format_impl(node: SyntaxNode, options: Options, context: Context) -> String {
...
 
 impl Ord for FormattedEntry {
     fn cmp(&self, other: &Self) -> cmp::Ordering {
-        self.key
-            .replace('\'', "")
-            .replace('"', "")
-            .cmp(&other.key.replace('\'', "").replace('"', ""))
+        self.cleaned_key()
+            .split('.')
+            .cmp(other.cleaned_key().split('.'))
```
ia0, line 397: No need to split.

ia0 (changes requested):

bebert64:
I apologize, I realize my previous message was far from clear :smiling_face_with_tear:  Thanks to your explanations, I now see how your solution provides consistency, but I would prefer the solution where that ordering is also consistent with the "classical" alphabetical order (`foo` before `foo-bar`). Since it's comparing iterator, I don't think there is a downside in performance, and believe it's probably closer to what most people would expect. Would that be okay with you?

ia0:
Yes, both were fine with me. So let's modify the `cleaned_key` to be `Vec<String>`.

bebert64:
No problem, thanks, I'll push the Vec modification in a min. For my curiosity (still learning :) ): wouldn't that require one more allocation than just splitting in place every time? Do you prefer it because it's easier to read at the (arguably very low) cost of just one additional allocation?

ia0:
Yes, there's one more allocation, but this is not high performance code (as you might have seen by the previous code allocating strings on every comparison) so it's not a concern. However readability and correctness by construction is more important here. In particular the fact that the comparison key exists in only one place is the key factor. Even without the comparison key being cached I would have been fine (i.e. `fn cmp_key(&self) -> Vec<String>` being called at each comparison for both sides).

ia0 (approved):
Looks good to me modulo the import formatting.

bebert64:
Thanks for the explanations, very clear and makes a lot of sense!

crates/taplo/src/formatter/mod.rs
```diff
@@ -3,17 +3,20 @@
...
+        util::overlaps,
+    },
+    once_cell::unsync::OnceCell,
+    rowan::{GreenNode, NodeOrToken, TextRange},
+    std::{
+        cmp,
+        iter::{repeat, FromIterator},
+        ops::Range,
+        rc::Rc,
+    },
```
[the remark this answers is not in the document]
bebert64: Absolutely, sorry about that (it's my auto-formating + not re-reading all changes, which was a mistake...)

ia0 (approved):
Thanks for the PR!

crates/taplo/src/formatter/mod.rs
```diff
@@ -358,36 +359,42 @@ fn format_impl(node: SyntaxNode, options: Options, context: Context) -> String {
 struct FormattedEntry {
     syntax: SyntaxElement,
     key: String,
+    // This field is used to cache the "cleaned" version of the key and should only
+    // be accessed through the `cleaned_key` helpers method.
+    private_cleaned_key: OnceCell<Vec<String>>,
```
panekj, line 364:
```suggestion
    /// This field is used to cache the "cleaned" version of the key and should only
    /// be accessed through the `cleaned_key` helpers method.
    cleaned_key: OnceCell<Vec<String>>,
```

panekj (changes requested):

crates/taplo/src/formatter/mod.rs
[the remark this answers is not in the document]
ia0: This got formatted again.

crates/taplo/src/formatter/mod.rs
```diff
@@ -361,14 +363,14 @@ struct FormattedEntry {
     key: String,
     // This field is used to cache the "cleaned" version of the key and should only
     // be accessed through the `cleaned_key` helpers method.
-    private_cleaned_key: OnceCell<Vec<String>>,
+    cleaned_key: OnceCell<Vec<String>>,
```
ia0, line 366: I think the idea was to also make the comment a documentation by using `///` instead of `//` (even though it's a private field; it's possible to generate documentation using `--document-private-items` and rust-analyzer always uses them).

ia0 (approved):

crates/taplo/src/formatter/mod.rs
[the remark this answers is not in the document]
bebert64: Omg... sorry again... :(

ia0 (approved):

panekj (approved):

bebert64:
Out of curiosity, how does the release process work ? I've seen that the commit has been released as part of [release-taplo-0.13.0](https://github.com/tamasfe/taplo/releases/tag/release-taplo-0.13.0), and I was wondering how it translates in terms of the vscode extension. Thanks for your insights

PR #521, opened 2023-12-26
Toggle off default `clap` dependency for `prettydiff`

CosmicHorrorDev:
(Context: Sniffing out different sources of old `clap` versions that are still getting downloaded)

`prettydiff` is one of those crates that also doubles as a CLI binary with a default `cli` feature for all the CLI bits. This is their only default feature and it pulls in `structopt` (which uses old clap v2.0) and `prettytable-rs`


ia0 (approved):
Thanks!

panekj (approved):

Issue #390, opened 2023-03-27
Array always auto collapses even when configured not to


cquick01:
I have an array defined like this

```toml
[build-system]
requires = [
    "poetry-core>=1.0.0",
    "setuptools",
    "setuptools_scm[toml]"
]
build-backend = "poetry.core.masonry.api"
```

But every time I save the file it reformats and collapses the array

```toml
[build-system]
requires = [ "poetry-core>=1.0.0", "setuptools", "setuptools_scm[toml]" ]
build-backend = "poetry.core.masonry.api"
```

even though I _explicitly_ configure as

```json
{
    "evenBetterToml.formatter.arrayAutoCollapse": false,
    "evenBetterToml.formatter.arrayAutoExpand": true,
}
```

Why is my array auto collapsing when I am explicitly configuring to auto expand? How do I fix this behavior? I've never had to configure this extension before and have modified these toml files numerous times before and they've never been reformatted this way.

---

Most of these options don't seem to have any effect. `evenBetterToml.formatter.arrayTrailingComma`, `evenBetterToml.formatter.compactEntries`, `evenBetterToml.formatter.indentEntries`, `evenBetterToml.formatter.reorderKeys` from what I've tried.

taranlu-houzz:
I am also running into this issue currently.

ia0:
> @tamasfe Is this project still active?

Yes, the project is alive but in maintenance mode for now, see #403.

> I am also running into this issue currently.

We didn't go through the issues backlog yet and probably only will based on activity like yours. I'm currently on vacations without access to a computer. However when I'm back in 2 weeks I'll definitely look at this issue, try to reproduce, and investigate. This will happen some time in the second half of June.

ia0:
(Sorry for the very long delay, I'm still going through my backlog.)

I tried to reproduce and cannot. Could you try again? Here's the steps I followed:
- Create an empty directory.
- Create `pyproject.toml` in that directory with the following (deliberately ill-formatted) content:
    ```toml
    [build-system]
    requires =   [
      "poetry-core>=1.0.0", "setuptools",
         "setuptools_scm[toml]",
    ]
    build-backend = "poetry.core.masonry.api"
    ```
- Create `taplo.toml` in that directory with the following content:
    ```toml
    [formatting]
    array_auto_collapse = false
    ```
- Run `taplo fmt pyproject.toml`.
[...]

PR #526, opened 2024-01-01
Clarify that array_auto_expand is conditional

clarkf:
Addresses part of #390.

The `array_auto_expand` formatter configuration option will prefer to leave arrays on a single line until they exceed the configured column length. The documentation didn't make this caveat clear, and lead some users to expect that all arrays would be expanded regardless.

This does not modify behavior at all, just documents it.


ia0 (approved):
Thanks for the PR!

Issue #542, opened 2024-02-01
`latest` release doesn't contain precompiled binaries


kziemianek:
The links from documentation stopped working since last release.

https://taplo.tamasfe.dev/cli/installation/binary.html
example: https://github.com/tamasfe/taplo/releases/latest/download/taplo-linux-x86_64.gz .

We switched from `latest` to `0.8.1` in order to get binaries and use them in CI. 
Is it possible to fix release so we can switch back to `latest`?

JounQin:
Sorry about this, I'll try to fix this issue.

kziemianek:
> Sorry about this, I'll try to fix this issue.

All good, thanks! 👍🏼

JounQin:
Sorry, this is because the release is published manually without CI. I'm not quite sure how to re-release, I'll try to remove the tag and recreate them manually.

JounQin:
<img width="1380" alt="image" src="https://github.com/tamasfe/taplo/assets/8336744/d6524855-4b4d-42ab-981d-ba1e9b3eafa8">

```log
[Invalid workflow file: .github/workflows/releases.yaml#L268](https://github.com/tamasfe/taplo/actions/runs/7742605055/workflow)
The workflow is not valid. .github/workflows/releases.yaml (Line: 268, Col: 19): Unrecognized named-value: 'env'. Located at position 1 within expression: env.RELEASE_VERSION .github/workflows/releases.yaml (Line: 270, Col: 19): Unrecognized named-value: 'env'. Located at position 1 within expression: env.RELEASE_VERSION
```

tamasfe:
Seems to be https://github.com/actions/runner/issues/1189 I have no idea why/how it worked before and what changed.

maresb:
See https://github.com/tamasfe/taplo/issues/535#issuecomment-1926446147. @nikaro writes:

> Yep, looks like the release workflow is broken: https://github.com/tamasfe/taplo/actions/runs/7742605055/workflow#L268
And it seems to be my fault 😬 the `env.RELEASE_VERSION` variable is not available in the matrix.

nikaro:
My bad, I'm working on a PR.

PR #547, opened 2024-02-05
ci(releases): fetch tools versions in its own job

nikaro:
Fixes: #542 

In #496 i introduced a bug in the release process by referencing a not yet defined environment variable `env.RELEASE_VERSION` in the matrix.
Causing this failure: https://github.com/tamasfe/taplo/actions/runs/7742605055/workflow#L268

This can be fixed by extracting the version fetching steps in its own `get_version` job and make jobs that require it to depend on it. Then instead of referencing an environment variable we use the `get_version` job outputs.

I also added the `workflow_dispatch` trigger to allow running the release workflow manually. So that it should be possible to re-run it to fix the releases, using the desired tag as the git ref:
Capture d’écran 2024-02-05 à 11 41 55

I've created a dummy repository to test the workflow logic (by fetching and echoing the versions) and it works fine: https://github.com/nikaro/github-actions-testing/actions/workflows/releases-taplo.yaml

* run for the `release-taplo-cli-*` tags: https://github.com/nikaro/github-actions-testing/actions/runs/7783089054
* run for the vscode extension tag: https://github.com/nikaro/github-actions-testing/actions/runs/7783089031

### DIFF
--- a/.github/workflows/releases.yaml
+++ b/.github/workflows/releases.yaml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - release-*
+  workflow_dispatch:
 
 # TODO: github doesn't support regex matching in if expressions for some reason,
 # so these jobs all expect zero-versioned release tags...
@@ -21,11 +22,28 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           wait-interval: 10
 
+  get_version:
+    runs-on: ubuntu-latest
+    needs: ["wait_for_ci"]
+    outputs:
+      cli: ${{ steps.cli.outputs.version || '' }}
+      vscode-ext: ${{ steps.vscode-ext.outputs.version || '' }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Retrieve release version
+        if: startsWith(github.ref, 'refs/tags/release-taplo-cli-0')
+        run: echo "version=$(echo ${GITHUB_REF#refs/tags/release-taplo-cli-})" >> $GITHUB_OUTPUT
+        id: cli
+      - name: Retrieve release version
+        if: startsWith(github.ref, 'refs/tags/release-even-better-toml-')
+        run: echo "version=$(echo ${GITHUB_REF#refs/tags/release-even-better-toml-})" >> $GITHUB_OUTPUT
+        id: vscode-ext
+
   cargo_publish_taplo:
     name: Cargo publish taplo
     runs-on: ubuntu-latest
     needs: ["wait_for_ci"]
-    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/release-taplo-0')
+    if: startsWith(github.ref, 'refs/tags/release-taplo-0')
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust toolchain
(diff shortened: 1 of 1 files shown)


JounQin:
Thanks for working this!

.github/workflows/releases.yaml
```diff
@@ -467,22 +482,16 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/download-artifact@v3
-
-      - name: Retrieve release version
```
JounQin, line 471: Can we keep this step?

JounQin (approved):

JounQin:
https://github.com/tamasfe/taplo/releases/tag/release-taplo-cli-0.9.1

@nikaro It seems tags are not released automatically anymore?

JounQin:
> [Invalid workflow file: .github/workflows/releases.yaml#L268](https://github.com/tamasfe/taplo/actions/runs/7742605055/workflow)
The workflow is not valid. .github/workflows/releases.yaml (Line: 268, Col: 19): Unrecognized named-value: 'env'. Located at position 1 within expression: env.RELEASE_VERSION .github/workflows/releases.yaml (Line: 270, Col: 19): Unrecognized named-value: 'env'. Located at position 1 within expression: env.RELEASE_VERSION

nikaro:
@JounQin the changes i did in this PR was to fix this issue specifically. But if you run the workflow with a tag prior to the PR i think that it still use the previous buggy version of the workflow.

You may need create a new one or override the existing one to point to more recent commit (on or after the workflow fix).

JounQin:
@nikaro So the release must be manual triggered now?

nikaro:
No, the trigger is still on tags pushes:

```yaml
on:
  push:
    tags:
      - release-*
```

JounQin:
https://github.com/tamasfe/taplo/tags

@nikaro They're not released automatically. (created 3 days ago)

nikaro:
Ok, i think i understand what's hapenning.

The `Continuous integration / Tag Releases` did run correctly, pushing new tags: https://github.com/tamasfe/taplo/actions/runs/7920213073/job/21622733377

But since @JounQin is the author here, it looks like the GitHub Personal Access Token of @tamasfe (in `secrets.WORKFLOW_PAT`) is not accessible. Thus it is empty and/or falls back on the "default" one, `GITHUB_TOKEN`, which cannot trigger another workflow (cf. [doc](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow)).

At least that's the only explanation i can think of. Because the Release workflow trigger didn't changed on this part it shouldn't behave differently than before. And i can (sort of) reproduce it:
- [default token auto tags pushed](https://github.com/nikaro/github-actions-testing/actions/runs/7959037182) -> release workflow not triggered
- [use of a custom PAT](https://github.com/nikaro/github-actions-testing/actions/runs/7959122939) -> [release workflow is run](https://github.com/nikaro/github-actions-testing/actions/runs/7959126144)

Did you try to run the workflow manually targeting the tags? In the meantime this could at least let us know if the Release workflow itself is working fine now.

nikaro:
Or maybe the `secrets.WORKFLOW_PAT` has been removed, or the PAT it references is expired? Or something like that...

ia0:
Thanks for the debugging work @nikaro !

@JounQin I guess you're the only one who can check the Settings page of the repository to check if the secrets are present. Otherwise, it's probably simpler to let @tamasfe push the release (once prepared). What do you think?

JounQin:
I don't have the permission to check the `Settings` page. 🤣

ia0:
Ah ok :D that solves it then. I think it's better if @tamasfe does the releases then, although we can still prepare them to minimize work.

Issue #535, opened 2024-01-20
Feature request: pre-commit hook for formatter


maresb:
It would be great if there were a [pre-commit hook](https://pre-commit.com/hooks.html) for the TOML formatter. My current go-to formatter is `pretty-format-toml`, but this suffers from a fairly serious issue that [comments are sometimes deleted](https://github.com/pappasam/toml-sort/issues/59). If this were implemented, I think it could easily become the best formatter with a pre-commit hook.

ia0:
Thanks for the suggestion! I guess this needs both the creation of some files in this repo and a PR for the pre-commit repository. This project is currently passively maintained, so only bugs are fixed. New functionalities need a PR contribution. Feel free to open one if you get the chance.

nikaro:
@maresb you can use mine if want: https://github.com/nikaro/taplo-pre-commit

Beware that the Docker hooks don't work on arm64 machines, until the next release of Taplo, cf. [this issue](https://github.com/nikaro/taplo-pre-commit/issues/2). The native Rust ones are fine, but the build will take a lot of time the first time you run it (after that it should be in your cache so it will be ok).

Feel free to reuse it to open a PR here if you want.

maresb:
Thanks @nikaro, I just now got around to trying this out. My experience is that it was going very slowly during the initial install. My suspicion is that pre-commit was downloading a rust toolchain in the background, and that's why it's so slow.

I was wondering what's the purpose of your Docker hooks, but if I understood the above correctly then it saves on all the compile time. Running Docker from pre-commit seems pretty crazy to me.  I wish it could just download binaries and use those. Unfortunately I'm not familiar with rust development, so I have no sense of the challenges of doing this.

Speaking of which, there are supposed to be [binary releases](https://taplo.tamasfe.dev/cli/installation/binary.html) but it seems like the links are broken and no binary artifacts are being generated by the GitHub workflows. Should I open a separate issue for that?

nikaro:
> Thanks @nikaro, I just now got around to trying this out. My experience is that it was going very slowly during the initial install. My suspicion is that pre-commit was downloading a rust toolchain in the background, and that's why it's so slow.

You are right that's what's happening: downloading rust toolchain and compiling. And that's why the `-docker` hook is here, to serve as a workaround to avoid having to compile (but it indeed requires Docker...).

> I wish it could just download binaries and use those. Unfortunately I'm not familiar with rust development, so I have no sense of the challenges of doing this.

[...]

PR #549, opened 2024-02-07
Build wheels for Python package

redeboer:
As discussed in https://github.com/ComPWA/mirrors-taplo/issues/13#issuecomment-1930749688, here is an attempt to build Python package wheels with [Maturin](https://github.com/PyO3/maturin). That would make it way easier for the Python community to integrate Taplo into their workflow and would also make it possible to host a fast Taplo pre-commit hook (https://github.com/tamasfe/taplo/issues/535).

A test `taplo` pip package is available here:
https://pypi.org/project/taplo-test

And you can use this as a test pre-commit hook:

```yaml
  - repo: https://github.com/redeboer/taplo-pre-commit
    rev: v0.9.1rc1
    hooks:
      - id: taplo
```

### OpenSSL dependency
Building wheels for MacOS and Windows was easy, but I ran into trouble with Linux because of the apparent need for OpenSSL. I thought that OpenSSL is not a dependency anymore (https://github.com/tamasfe/taplo/pull/302), but it is still [pulled in indirectly](https://github.com/tamasfe/taplo/blob/66cc4e402433f61c22adff98f68069b42f92a268/Cargo.lock#L1400-L1452) through `reqwest`:
  https://github.com/tamasfe/taplo/blob/4bf7b536bcf03369ec9318000d3e88e1bbf994ac/crates/taplo-cli/Cargo.toml#L31-L33

It seems that `reqwest` is used in several of the taplo crates, so I presume we have to deal with OpenSSL as a dependency. Manylinux does not offer OpenSSL however, which [resulted in broken wheels](https://github.com/redeboer/taplo/actions/runs/7814055668/job/21314888768). I therefore added [`openssl` as an optional dependency](https://github.com/redeboer/taplo/blob/66cc4e402433f61c22adff98f68069b42f92a268/crates/taplo-cli/Cargo.toml#L31) with the vendored flag, i.e. `cargo build --features vendored-openssl` (open to suggestions for a better name).[^lock-update]
[^lock-update]: 55c7e96 upgrades the lock files to keep the diff of 2ef0d97 small; you can revert that lock update if preferred.

Even though that solution solution [seems to work](https://github.com/redeboer/taplo/actions/runs/7775977425)[^package-job], it only 'bakes' openssl into the wheels for Linux x86_64: the 32-bit version is [still built](https://github.com/redeboer/taplo/blob/72f4b60ddd64ddf58af8030a7fed957192f17e16/.github/workflows/python.yml#L35-L44) with shared openssl lib, because its manylinux [still runs into problems](https://github.com/redeboer/taplo/actions/runs/7814054400/job/21314487625) building with vendored OpenSSL. And I haven't managed to test whether those wheels actually work.
[^package-job]: This is the job that published the [`taplo-test`](https://pypi.org/project/taplo-test) package.

> [!IMPORTANT]
> Who can help test whether the x86 wheels work? You can try the [`taplo-test`](https://pypi.org/project/taplo-test) package or the [artifacts from the job](https://github.com/redeboer/taplo/actions/runs/7775977425).

### PyPI release
I'm happy to help maintain this repository (https://github.com/tamasfe/taplo/issues/403), especially the Python release. For now, however, we need to somehow share a [`MATURIN_PYPI_TOKEN`](https://github.com/tamasfe/taplo/blob/72f4b60ddd64ddf58af8030a7fed957192f17e16/.github/workflows/python.yml#L158) for the workflow. **Who of the Taplo maintainers is also on PyPI?**

### Pre-commit hook
To address https://github.com/tamasfe/taplo/issues/535, it will be necessary to put the `.pre-commit-hooks.yaml` under a different config (see example repo [here](https://github.com/redeboer/taplo-pre-commit)). The reason is that (1) it speeds up the pre-commit clone and (2) this repo already has a `pyproject.toml` for the Maturin build. This would be an argument for creating a 'Taplo organization' https://github.com/tamasfe/taplo/issues/403#issuecomment-1913342635. [^vscode]

[^vscode]: An organization would also make it easier to extract the [VSCode extension](https://github.com/tamasfe/taplo/tree/1aed562/editors/vscode) to separate repo.


.github/workflows/python.yml
```diff
@@ -0,0 +1,163 @@
...
+      - if: matrix.target == 'x86_64'
+        name: Build wheels (static openssl)
+        uses: PyO3/maturin-action@v1
+        with:
+          target: ${{ matrix.target }}
+          args: --release --out dist --features vendored-openssl
+          sccache: "true"
+          manylinux: auto
+          before-script-linux: |
+            yum install -y perl-IPC-Cmd
```
nikaro, line 36: Shouldn't it be gated with something like `command -v yum &>/dev/null && yum install [...]`? As `yum` is only available on RHEL-like distros.
And then, wouldn't it be needed to have the same for apt/apk/dnf/pacman/whatever? :-/

nikaro: Ok nevermind, this is used only in the build process... i thought it was like post-install command of the wheel package.

flying-sheep:
Is there anything blocking this from being merged?

flying-sheep:
Yeah, I think there’s no way around bundling at the moment, and having only 64 bit wheels is probably fine for now.

If someone wants 32 bit, they can start complaining I guess.

bollwyvl:
This would be lovely! Is it anticipated the PyPI package would carry the `taplo-cli` version number, or something else? For downstreams that already have a `taplo` package, e.g. [`conda-forge`](https://github.com/conda-forge/taplo-feedstock/issues/5), would there be a preferred name for the python package when not installed from PyPI, e.g. `python-taplo`?

A big thing for making this an easily-adoptable tool for the broader python ecosystem would be if `taplo` (even the existing fat binary CLI) could be coaxed into looking in `pyproject.toml#/tool/taplo` to reduce dotfile proliferation.

bollwyvl:
> conda-forge feedstock would just pull from PyPI

For _most_ python packages, yes: we use `pip install` of expanded `.tar.gz` sdists from `files.pythonhosted.org`. Wheels are used in a _very_ small number of cases, sometimes due to egregious abuse of magic PEP517 `[build-system]`s. As a `conda-forge` maintainer, I _really_ don't want `pip` installing `nodejs` from... wherever during my build. Similarly, I wouldn't want `pre-commit` pulling in... whatever... from... wherever. Hence, I'm more interested in the intentional case of declaring an optional dependency in `pyproject.toml` (a la #603), and will happily let `pre-commit` take care of itself.

Back on `taplo-???` on `conda-forge`: we probably wouldn't touch PyPI at all. As a `rust` project, the canonical source (for the source) of `taplo` would likely continue to be the tagged GitHub archive, as it would contain whatever metadata is needed. We would patch some way to _not_ invoke `maturin` at all, and still leave `$PREFIX/site-packages/taplo-0.9.2.dist-info/` in a state that can appropriately fool `pip check`, one of the few tools that we have to validate the output of the galaxy of PEP517 build backends at scale.

bollwyvl:
To be clear: I am _not_ suggesting what `conda-forge` will need to do should impact what _this project_ should do on PyPI: `pip install taplo && taplo fmt pyproject.toml` feels very nice!

redeboer:
@nikaro @panekj Is there anything blocking this from being merged?
There still seems to be quite some interest in a fast pre-commit formatter for TOML files

panekj:
> Is there anything blocking this from being merged?

I don't understand why is it required for pre-commit tool and I'm not keen on supporting additional language ecosystem that isn't familiar to me.
Additionally it introduces a whole new workflow just to build a python package that isn't integrated with release workflow.

mrijken:
A Python variant is in a lot of use cases easier to install than a Rust variant and the number of possible users will grow with huge numbers when a Python variant is supported. I'm sure the community can and will help you to support the (small) Python part.

In our company we want to integrate a toml formatter in our ci/cd pipeline. Without a Python package, it is hard for us to use it.

flying-sheep:
> that isn't integrated with release workflow.

are you sure?

```yaml
    tags:
      - release-*cli-*
```

looks to me like the workflow is triggered by pushing to release branches, which seems like integration with the release process to me.

panekj:
> are you sure?

I'm sure
image

redeboer:
> Additionally it introduces a whole new workflow just to build a python package that isn't integrated with release workflow.

That's true, at the time I created a separate workflow, but over the past half year the existing workflows [were extended quite a bit](https://github.com/tamasfe/taplo/compare/e1938a8...d1d5a0b#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133dd).

I hope ffd444a is more agreeable to the intended release flow. What I like less about it, is that the Maturing build process is now only tested on releases.

> A Python variant is in a lot of use cases easier to install than a Rust variant and the number of possible users will grow with huge numbers when a Python variant is supported.

I agree, see also issues like #603 and [other issues related to `pyproject.toml`](https://github.com/tamasfe/taplo/issues?q=is%3Aissue+is%3Aopen+pyproject).

redeboer:
> What I like less about it, is that the Maturing build process is now only tested on releases.

Oww okay, nvm 😅 
https://github.com/tamasfe/taplo/actions/runs/10459526952/job/28964054546?pr=549
Fixed by 2ed2723

panekj:
This will also require a token configured in secrets and until project owner becomes active again, that's impossible to implement.

redeboer:
> This will also require a token configured in secrets and until project owner becomes active again, that's impossible to implement.

That's right, see PR description

> ### PyPI release
> I'm happy to help maintain this repository (#403), especially the Python release. For now, however, we need to somehow share a [`MATURIN_PYPI_TOKEN`](https://github.com/tamasfe/taplo/blob/72f4b60ddd64ddf58af8030a7fed957192f17e16/.github/workflows/python.yml#L158) for the workflow. **Who of the Taplo maintainers is also on PyPI?**

panekj:
> That's right, see PR description

I've seen that, I'm telling you why I won't merge it

FlavioAmurrioCS:
My two cents.

- The owner of the pypi project should be able to add @tamasfe as a owner
- @tamasfe should then remove the original owner. 
- @tamasfe should then remove all other tokens and create a project token
- finally add the new token as a repo secret. 

I believe the only owner of the pypi project should be the owner of taplo so that we only need to trust one person.

FlavioAmurrioCS:
Alternatively, you could try adding the project as a trusted publisher so that a token is not needed.

https://docs.pypi.org/trusted-publishers/adding-a-publisher/

s-weigand:
Two more cents

Using `trusted publisher` is the recommended way by PyPA and the default of [`pypa/gh-action-pypi-publish`](https://github.com/pypa/gh-action-pypi-publish).

redeboer:
Thanks for the tips/cents! ;)
Ah yes, I was overthinking with Maturin, you can indeed just use PyPI's recommended publish action.

I'm happy to transfer ownership of https://pypi.org/project/taplo/ to @tamasfe. Could you confirm which PyPI account to transfer it to?

.github/workflows/releases.yaml
```diff
@@ -436,3 +436,79 @@ jobs:
...
+          - os: windows
+            arch: aarch64
+          - os: ubuntu
+            arch: aarch64
+          - os: macos
+            arch: x86
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build sdist
+        if: matrix.os == 'ubuntu' && matrix.arch == 'x86_64'
```
panekj, line 459: todo: fix

redeboer: `sdist` is just the source code, so no need to compile that on different architectures. That's why it was in a separate job. It is not really required though.
https://docs.python.org/3.10/distutils/sourcedist.html

panekj: Yes, but the arch comparison was wrong and I was too tired to fix it yesterday

redeboer:
Awesome, thanks for merging!

danf-extrahop:
Didn't want to open a whole new issue or anything, but y'all should add `pip` installation instructions to the website/readme/whatever. I was scouring the Internet hoping to not have to add Cargo to my devcontainers just for taplo and initially didn't trust https://pypi.org/project/taplo/ since its maintainer was different from Github.

PR #554, opened 2024-02-13
Do not enable default-tls unconditionally

### LINKED ISSUE #553: Why was default-tls unconditionally enabled for reqwest?
I can't find the PR that merged https://github.com/tamasfe/taplo/commit/4be1d1eb9fba62ddd6cc7cf53058b6bce3d66c7c and the commit description doesn't provide a reason. @JounQin, was this pushed without PR? And what was the reason for this change? This unconditionally adds libssl3 as a dynamic library which didn't use to be the case.

ia0:
This reverts 4be1d1eb9fba62ddd6cc7cf53058b6bce3d66c7c and make it more convenient to build taplo-common without explicitly providing a feature.

Fixes #553

### DIFF
--- a/crates/taplo-cli/Cargo.toml
+++ b/crates/taplo-cli/Cargo.toml
@@ -35,7 +35,7 @@ schemars = "0.8"
 serde = "1"
 serde_json = "1"
 taplo = { version = "0.13.0", path = "../taplo", features = ["serde"] }
-taplo-common = { version = "0.5.0", path = "../taplo-common" }
+taplo-common = { version = "0.5.0", path = "../taplo-common", default-features = false }
 taplo-lsp = { version = "0.7.0", path = "../taplo-lsp", default-features = false, optional = true }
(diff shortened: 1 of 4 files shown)


JounQin (approved):

JounQin:
Package versions should be updated at the same time? Or would you work on another PR for releasing?

ia0:
> Package versions should be updated at the same time? Or would you work on another PR for releasing?

If I can avoid dealing with the release process I would prefer. I don't have too much time to invest. And actually before releasing it would probably be nice to update the `Cargo.lock` file to stop using yanked versions.

```
warning: package `ahash v0.7.6` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
warning: package `ahash v0.8.3` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
warning: package `crossbeam-channel v0.5.6` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
```

PR #556, opened 2024-02-14
chore: bump all packages versions

JounQin:
try to release with CI for precompiled binaries


crates/lsp-async-stub/Cargo.toml
```diff
@@ -3,7 +3,7 @@ authors = ["tamasfe"]
 description = "An LSP server stub for futures"
 edition = "2021"
 name = "lsp-async-stub"
-version = "0.6.2"
+version = "0.6.3"
```
ia0, line 6: I don't think this is needed. There were no changes since 4bf7b536bcf03369ec9318000d3e88e1bbf994ac (which AFAICT is the last release).

crates/taplo-cli/Cargo.toml
ia0, line 1: We also need to bump the `taplo-common` dependency version. Same for `taplo` and `taplo-lsp`.

crates/taplo-lsp/Cargo.toml
ia0, line 1: Same here, we need to bump `taplo-common` and `taplo`.

editors/vscode/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "even-better-toml",
   "displayName": "Even Better TOML",
   "description": "Fully-featured TOML support",
-  "version": "0.21.0",
+  "version": "0.20.0",
```
ia0, line 5: I'm not familiar with VS Code, but is this expected to go down?

js/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@taplo/cli",
-  "version": "0.7.0",
+  "version": "0.7.1",
```
ia0, line 3: I'm assuming those numbers are not expected to match those of `Cargo.toml`.

ia0 (commented):
Thanks!

Also, don't you think we should merge #510 first? I'm adding you as reviewer.

JounQin:
@ia0 I'd like to merge that PR as a minor release.

crates/lsp-async-stub/Cargo.toml
JounQin: We need to release on CI.

crates/taplo-cli/Cargo.toml
JounQin: AFAIK, the version strategy will work without change `Cargo.yaml`? (They have been updated in lock files)

editors/vscode/package.json
JounQin: Both `0.20.0` and `0.21.0` are not published actually, and I believe `0.21.0` is a typo.

js/cli/package.json
JounQin: They don't need to be matched.

crates/taplo-lsp/Cargo.toml
JounQin: As above, no need to change them in `Cargo.yaml`.

crates/taplo-cli/Cargo.toml
ia0: The `Cargo.lock` file is only relevant for `cargo install --locked`. It's true that `cargo install` would still work because it would use the most recent crate. However, for libraries like `taplo-lsp` (and actually `taplo-cli` too, because it's both a library and a binary) the `Cargo.lock` file is irrelevant.

I think it is safer to bump those dependencies (and actually, we might want to yank `taplo-common` 0.5.0 because an old `taplo{,-cli,-lsp}` using the new `taplo-common` would misbehave or fail to compile given an incompatible TLS feature).

crates/lsp-async-stub/Cargo.toml
ia0: Sounds good.

ia0:
> @ia0 I'd like to merge that PR as a minor release. HDYT?

Sounds good to me.

crates/taplo-cli/Cargo.toml
JounQin: > I think it is safer to bump those dependencies

Sounds reasonable to me. 👍

ia0 (approved):

PR #565, opened 2024-02-20
fix: allow building with `no-default-features`

### LINKED ISSUE #564: `no-default-features` is broken
Right now taplo-cli can't be installed without TLS because `no-default-features` is broken

Repro:
```
>cargo install --force --git https://github.com/tamasfe/taplo taplo-cli --no-default-features
error[E0433]: failed to resolve: could not find `Certificate` in `reqwest`
   --> crates/taplo-common/src/util.rs:136:25
    |
136 |             Ok(reqwest::Certificate::from_der(&buf)?)
    |                         ^^^^^^^^^^^ could not find `Certificate` in `reqwest`
    |
```

Daniel-Aaron-Bloom:
Fixes #564.

I just hid all the TLS inside a conditionally defined helper.

### DIFF
--- a/crates/taplo-cli/Cargo.toml
+++ b/crates/taplo-cli/Cargo.toml
@@ -35,7 +35,7 @@ schemars = "0.8"
 serde = "1"
 serde_json = "1"
 taplo = { version = "0.13.1", path = "../taplo", features = ["serde"] }
-taplo-common = { version = "0.5.1", path = "../taplo-common", default-features = false }
+taplo-common = { version = "0.5.1", path = "../taplo-common" }
 taplo-lsp = { version = "0.7.1", path = "../taplo-lsp", default-features = false, optional = true }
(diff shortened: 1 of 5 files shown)


ia0 (commented):
Looks good to me.

Could you also partially revert #554? (remove the default feature from `taplo-common` and remove the associated `default-features = false`, essentially only keeping the deletion of `default-tls` in the reqwest dependency of taplo-common) Thanks!

ia0 (approved):
Thanks a lot! That's a much better fix than mine. I didn't notice it was a change on reqwest side.

JounQin:
We'll wait for a while before  merging in case for new reviews.

Daniel-Aaron-Bloom:
Ping. 🙂

JounQin:
@Daniel-Aaron-Bloom Thanks for your contribution!

PR #573, opened 2024-03-03
Updated JSON schemas

alfawal:
Updated the JSON schemas to the latest version available at schemastore.org.


ia0:
Are those files even used? Taplo should download schemas from SchemaStore directly.

panekj:
Closing since taplo uses schemas from schemastore

PR #598, opened 2024-04-18
Update taplo to use workspace dependencies

panekj:
Fixes https://github.com/tamasfe/taplo/issues/628
Fixes https://github.com/tamasfe/taplo/issues/643


panekj:
I think this has marinated long enough, yolo

PR #609, opened 2024-05-07
Fix realign of comment makes debug assert fail

gaborbernat:
In this case, the two comments match up.

Discovered while doing https://github.com/gaborbernat/pyproject-fmt-rust


panekj:
Please mind that there are other maintainers as well.

panekj (approved):

gaborbernat:
Can we merge and land this pleaes then?

gaborbernat:
Can we land this, please? Thanks!

panekj:
It's already in HEAD, you can use patch in Cargo manifest if you need it

gaborbernat:
How do you do?

Issue #628, opened 2024-07-06
Build broken on latest nightly


scrabsha:
I hit the following error when installing `taplo`:

```
$ cargo install taplo-cli --locked
error[E0635]: unknown feature `stdsimd`
  --> /home/scrabsha/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ahash-0.7.6/src/lib.rs:33:42
   |
33 | #![cfg_attr(feature = "stdsimd", feature(stdsimd))]
   |                                          ^^^^^^^

```

Maybe this can be fixed by bumping one of the dependencies?

I'm using the following nightly:

```
$ rustc --version
rustc 1.81.0-nightly (524d806c6 2024-07-05)
```

panekj:
Is there any reason you can't use stable rust?

scrabsha:
None, nightly is just my default toolchain

I'm fine with you supporting only stable Rust (it's even better, actually!), but this should be documented.

scrabsha:
Can confirm it builds on stable. I'm going to update the taplo docs then :)

panekj:
> but this should be documented.

I don't think that us not supporting constantly changing highly experimental compiler versions is something we should pertain ourselves in documentation. After all one nightly might work, other might don't and the status changes constantly.

scrabsha:
fine. feel free to close this issue if you think the doc are good as is.

panekj:
I'll leave it open for now in case any other maintainer becomes active again and thinks otherwise

panekj:
Also a note about dep
> Maybe this can be fixed by bumping one of the dependencies?

Unfortunately it's `ahash` which brings circular dependency issues and we already had few times problems dealing with it, I'm currently unsure if that dependency leaf can be brought up on par with the other one (we have `0.7` and `0.8` `ahash` in lockfile) but I'll try looking into it at some point.

awnion:
Works without `--locked` so maybe bumping `.lock` would be a good idea

PR #639, opened 2024-07-28
Update almost all dependencies

dimbleby:
updates most dependencies

I've ducked a couple
- latest `lsp-types` is 0.97.0
- latest `logos` is 0.14.0

in both cases, upgrading taplo to the next available version introduces more breaking changes than I care to deal with


dimbleby:
fixes #628, presumably

PR #640, opened 2024-07-30
Move to `ignore` for ignoring files

### LINKED ISSUE #619: `taplo-cli` very slow at collecting files
Our project grew and `taplo` evolved. Trying `0.9.0` and `0.9.1` (unreleased, git) both exhibit a rather slow glob resolution and file collection. It hangs for minutes.

drahnr:
Closes #619

A pragmatic and practical fix to slow performance on directories with large `target` dirs et.al. (form _doesn't terminate_ to _milliseconds_ for 100GB+ `target` directory)

Also avoids a bunch of unnecessary `collect`ions and passes around an iterator instead, with the according file rename.

### DIFF
--- a/crates/taplo-cli/src/lib.rs
+++ b/crates/taplo-cli/src/lib.rs
@@ -1,6 +1,5 @@
 use anyhow::{anyhow, Context};
 use args::GeneralArgs;
-use itertools::Itertools;
 use std::{
     path::{Path, PathBuf},
     str,
@@ -13,6 +12,8 @@ use taplo_common::{config::Config, environment::Environment, util::Normalize};
 pub mod args;
 pub mod commands;
 pub mod printing;
+
+pub type EntryIter<'a, F> = std::iter::FilterMap<ignore::Walk, F>;
 
 pub struct Taplo<E: Environment> {
     env: E,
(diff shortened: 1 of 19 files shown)


crates/taplo-cli/src/lib.rs
```diff
@@ -111,33 +120,67 @@ impl<E: Environment> Taplo<E> {
...
-        Ok(files)
+        let cwd = cwd.to_path_buf();
+
+        let mut bldr = ignore::WalkBuilder::new(&cwd);
+        bldr.git_ignore(true)
+            .git_exclude(true)
+            .git_global(true)
+            .ignore(true)
+            .hidden(false)
+            .same_file_system(true);
```
findepi, line 141: I feel the same as you that obeying gitignore rules is likely a desired behavior and i personally have nothing against this becoming the default behavior (one day)

however, I agree with @panekj https://github.com/tamasfe/taplo/issues/619#issuecomment-2162838457 that this can be viewed as  a breaking change, so would be better to have this as an opt-in feature, at least for now.

Also, given how `ignore` API is designed -- _some ignore rules by default_ --  the walk builder should be configured with `standard_filters(false)` first, then enable what we want.

Question
is switching to `ignore` create sufficient to prune large target directly without recursing to it?

ForsakenHarmony:
Any chance this could get merged soon?
Quite frustrating having to wait 2 seconds every time I run taplo

(might need to be rebased)

drahnr:
I am happy to rebase if there is a clear intent to merge it

ForsakenHarmony:
cc @panekj

AH-Hammi:
What is preventing this feature from getting merged?
I mean if it helps it could be behind a global configuration flag?

Issue #353, opened 2022-11-18
Infinite recursion for completions


tamasfe:
Schemas are crawled for completions with a given depth limit. However composite schemas (e.g. allOf, anyOf) are currently not counted as child schemas and will not count towards the schema depth. Thus it is possible to cause a stack overflow by mutually referencing schemas that are all composite schemas themselves (e.g.  `Target` in https://raw.githubusercontent.com/metalbear-co/mirrord/main/mirrord-schema.json).

The fix itself is trivial, just treat composite schemas as children. However this will change the semantics of the user-provided limit that restricts the length of the suggested dotted keys.

To handle this properly:
- We should treat the current depth as some internal hard-limit to prevent infinite recursions.
- The user-defined limit will apply after all schemas are collected, thus will not affect performance anymore but purely UX.

PR #644, opened 2024-08-11
[taplo-common] fix infinite recursion with composed allOfs

sunshowers:
Hi there!

I was trying to write a schema for [nextest's configuration](https://nexte.st/docs/configuration/), and I ran into a call stack exhaustion issue. It turned out to be because an `allOf` wasn't getting removed even though there was an attempt to do so, leading to infinite recursion.

I couldn't find any tests for schemas so I added some. Please let me know if I can rearrange this to your liking somehow, thanks!

(And once this is landed a release would be lovely -- this is a hard blocker for nextest's schema at the moment.)

### DIFF
--- a/crates/taplo-common/src/schema/mod.rs
+++ b/crates/taplo-common/src/schema/mod.rs
@@ -543,9 +543,12 @@ impl<E: Environment> Schemas<E> {
         if let Some(all_ofs) = schema["allOf"].as_array() {
             if !all_ofs.is_empty() && composed {
                 let mut schema = schema.clone();
-                if let Some(obj) = schema["allOf"].as_object_mut() {
-                    obj.remove("allOf");
-                }
+                // Remove the allOf because we're resolving in this spot -- not doing so will cause
+                // infinite recursion.
+                schema
+                    .as_object_mut()
+                    .expect("schema is an object")
+                    .remove("allOf");
 
                 let mut merged_all_of = Value::Object(serde_json::Map::default());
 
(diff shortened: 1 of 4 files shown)


crates/taplo-common/src/schema/mod.rs
```diff
@@ -543,9 +543,12 @@ impl<E: Environment> Schemas<E> {
...
-                    obj.remove("allOf");
-                }
+                // Remove the allOf because we're resolving in this spot -- not doing so will cause
+                // infinite recursion.
+                schema
+                    .as_object_mut()
+                    .expect("schema is an object")
+                    .remove("allOf");
 
                 let mut merged_all_of = Value::Object(serde_json::Map::default());
```
sunshowers, line 554: I noticed below in line 569, and above in line 512/518/525, that `depth` wasn't getting subtracted by 1. Is this expected? Naively I'd expect `depth` to be subtracted here.

sunshowers: Ah, it looks like https://github.com/tamasfe/taplo/issues/353 is related.

DavisVaughan:
This one would be really nice to merge! Looks to me like it is just fixing a small bug and making the intended behavior actually happen. It did cause parsing of ruff's toml to break (https://github.com/astral-sh/ruff/issues/15978) which they ended up working around but it would be nice to not have to.

sunshowers:
Ah looks like the project is active again -- apologies for pinging you directly @panekj but I'm wondering if this PR can be merged. The bug is pretty clear and it's a hard blocker for nextest's schema.

sunshowers:
Closing and re-doing this after rebasing.

PR #664, opened 2024-08-30
improve error locations for unexpected properties

cdmistman:
## Why
At replit we use taplo for providing lsp support for our [`.replit` configuration file](https://docs.replit.com/replit-workspace/configuring-repl#replit-file) in Repls. We love Taplo!

However, when an unexpected property appears that's invalid according to `additionalProperties: false` from the corresponding JSON schema, the LSP highlights the whole file!

for example, making invalid changes to this valid `.replit` file i used for a [throwaway Repl I made when debugging this issue](https://replit.com/t/replit/onreplit/repls/japlo/view):
```toml
run = "cargo run -- /etc/replit/dotreplit.schema.json dotreplit.json"
modules = ["rust-stable"]

[nix]
channel = "stable-24_05"
```

results in this UX:
image

note in this screenshot that the diagnostics are collated to the first diagnostic range, which is the whole file:
image
image
image

those error spans aren't the most helpful, especially that 2nd screenshot 😬

## What changed

1. updated `taplo::dom::Node::text_ranges` method to accept a new argument `include_children: bool` which conditionally recurses into children when the value is true. `true` was the default behavior prior to this change
2. created new `taplo_common::schema::NodeValidationError::text_ranges` method to conditionally pass `false` to the above method.
   - right now, only `jsonschema::error::ValidationErrorKind::AdditionalProperties` errors result in this value being `false`, but it may be desired that other error kinds should result in this value being `false`
   - i wanted to minimize changing behavior in case this resulted in unexpected behavior
3. updated `lint` and `lsp` commands to use the new `NodeValidationError::text_ranges` method instead of `Node::text_ranges`
4. all other calls to `Node::text_ranges` pass `true` as the parameter to minimize unexpected behavior

this results in this new behavior:
image
image

### DIFF
--- a/crates/taplo-cli/src/printing.rs
+++ b/crates/taplo-cli/src/printing.rs
@@ -123,7 +123,7 @@ impl<E: Environment> Taplo<E> {
         let mut out_diag = Vec::<u8>::new();
         for err in errors {
             let msg = err.error.to_string();
-            for text_range in err.node.text_ranges() {
+            for text_range in err.text_ranges() {
                 let diag = Diagnostic::error()
                     .with_message(err.error.to_string())
                     .with_labels(Vec::from([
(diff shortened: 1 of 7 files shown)


cdmistman:
note in the new-behavior screenshots that only the value is underlined with the error. i'd like this to underline the full entry, but i'll figure that out later

panekj:
Thanks!

PR #668, opened 2024-09-03
Add uv.lock as being identified automatically as TOML

### LINKED ISSUE #667: Add `uv.lock` as being identified automatically as TOML
https://github.com/astral-sh/uv has an `uv.lock` file that is TOML. It would be nice if this extension adds the filename `uv.lock` as being automatically identified as being encoded as TOML.

thernstig:
Fixes #667


bryan-hunted:
I'm not a maintainer here so my opinion doesn't really matter. That said, I don't particularly want to have an external formatter edit files that are automatically generated. It's bound to cause back-and-forth between the tool output and the formatter output and thus make file diffs noisier. I'd prefer if this was not added as a default.

thernstig:
@bryan-hunted this is not about editting it, but to give it colorization etc. Obviously if you do open and save it, it gets edited. But that is true for anyone who anyway marks this file as a YAML file.

So the issue is about identifiying it as TOML correctly.

PR #730, opened 2025-01-09
Update jsonschema to 0.28.1, support drafts 2019-09 and 2020-12

### LINKED ISSUE #497: Discrepancy in validation with JSON Schema draft 2020-12

uncenter:
Closes #497, jsonschema now supports both of drafts (https://github.com/tamasfe/taplo/issues/497#issuecomment-2381386773).

### DIFF
--- a/crates/taplo-common/src/schema/mod.rs
+++ b/crates/taplo-common/src/schema/mod.rs
@@ -5,7 +5,9 @@ use async_recursion::async_recursion;
 use futures::{stream::FuturesUnordered, StreamExt};
 use itertools::Itertools;
 use json_value_merge::Merge;
-use jsonschema::{error::ValidationErrorKind, JSONSchema, SchemaResolver, ValidationError};
+use jsonschema::{
+    error::ValidationErrorKind, ReferencingError, Retrieve, ValidationError, Validator,
+};
 use parking_lot::Mutex;
 use regex::Regex;
 use serde_json::Value;
@@ -50,7 +52,7 @@ pub struct Schemas<E: Environment> {
     associations: SchemaAssociations<E>,
     concurrent_requests: Arc<Semaphore>,
     http: reqwest::Client,
-    validators: Arc<Mutex<LruCache<Url, Arc<JSONSchema>>>>,
+    validators: Arc<Mutex<LruCache<Url, Arc<Validator>>>>,
     cache: Cache<E>,
 }
 
@@ -125,7 +127,7 @@ impl<E: Environment> Schemas<E> {
 
     async fn validate_impl(
         &self,
-        validator: &JSONSchema,
+        validator: &Validator,
         value: &Value,
     ) -> Result<Vec<ValidationError<'static>>, anyhow::Error> {
         // The following loop is required for retrieving external schemas.
@@ -134,52 +136,53 @@ impl<E: Environment> Schemas<E> {
         // a validation path that requires it, so we might have to loop many times
         // to fully validate according to a schema that has many nested references.
         loop {
-            match validator.validate(value) {
-                Ok(()) => return Ok(Vec::new()),
-                Err(errors) => {
-                    let errors: Vec<_> = errors
-                        .map(|err| ValidationError {
-                            instance: Cow::Owned(err.instance.into_owned()),
-                            kind: err.kind,
-                            instance_path: err.instance_path,
-                            schema_path: err.schema_path,
+            let errors = validator
+                .iter_errors(value)
+                .map(|err| ValidationError {
+                    instance: Cow::Owned(err.instance.into_owned()),
+                    kind: err.kind,
+                    instance_path: err.instance_path,
+                    schema_path: err.schema_path,
+                })
+                .collect::<Vec<_>>();
+
+            // We check whether there were any external schema errors,
+            // and retrieve the schemas accordingly.
+            let mut external_schema_requests: FuturesUnordered<_> = errors
+                .iter()
+                .filter_map(|err| {
+                    if let ValidationErrorKind::Referencing(ReferencingError::Unretrievable {
+                        uri,
+                        ..
+                    }) = &err.kind
+                    {
+                        Some(async {
+                            let uri = Url::parse(uri)?;
+                            let value = self.load_schema(&uri).await?;
+                            drop(self.cache.store(uri, value));
+                            Result::<(), anyhow::Error>::Ok(())
                         })
-                        .collect();
-
-                    // We check whether there were any external schema errors,
-                    // and retrieve the schemas accordingly.
-                    let mut external_schema_requests: FuturesUnordered<_> = errors
-                        .iter()
-                        .filter_map(|err| {
-                            if let ValidationErrorKind::Resolver { url, .. } = &err.kind {
-                                Some(async {
-                                    let value = self.load_schema(url).await?;
-                                    drop(self.cache.store(url.clone(), value));
-                                    Result::<(), anyhow::Error>::Ok(())
-                                })
-                            } else {
-                                None
-                            }
-                        })
-                        .collect();
-
-                    // There are no external schemas to retrieve,
-                    // return the errors as-is.
-                    if external_schema_requests.is_empty() {
-                        drop(external_schema_requests);
-                        return Ok(errors);
+                    } else {
+                        None
                     }
+                })
+                .collect();
 
-                    // Retrieve external schemas, and return on the first failure.
-                    while let Some(external_schema_result) = external_schema_requests.next().await {
-                        external_schema_result?;
-                    }
+            // There are no external schemas to retrieve,
+            // return the errors as-is.
+            if external_schema_requests.is_empty() {
+                drop(external_schema_requests);
+                return Ok(errors);
+            }
 
-                    // Try validation again, now with external schemas
-                    // resolved and cached.
-                    continue;
-                }
-            };
+            // Retrieve external schemas, and return on the first failure.
+            while let Some(external_schema_result) = external_schema_requests.next().await {
+                external_schema_result?;
+            }
+
+            // Try validation again, now with external schemas
+            // resolved and cached.
+            continue;
         }
     }
 
@@ -217,7 +220,7 @@ impl<E: Environment> Schemas<E> {
         Ok(schema)
     }
 
-    fn get_validator(&self, schema_url: &Url) -> Option<Arc<JSONSchema>> {
+    fn get_validator(&self, schema_url: &Url) -> Option<Arc<Validator>> {
         if self.cache().lru_expired() {
             self.validators.lock().clear();
         }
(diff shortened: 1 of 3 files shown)


crates/taplo-common/src/schema/mod.rs
```diff
@@ -705,11 +707,10 @@ impl NodeValidationError {
...
                     _ => return Err(anyhow!("invalid key")),
                 },
-                jsonschema::paths::PathChunk::Index(idx) => {
-                    node = node.try_get(*idx).map_err(|_| anyhow!("invalid index"))?;
-                    keys = keys.join(*idx);
+                jsonschema::paths::LocationSegment::Index(idx) => {
+                    node = node.try_get(idx).map_err(|_| anyhow!("invalid index"))?;
+                    keys = keys.join(idx);
                 }
-                jsonschema::paths::PathChunk::Keyword(_) => {}
```
uncenter, line 712: Not quite sure what this was but I couldn't figure out what the equivalent was based on the release notes and migration guide of jsonschema.

crates/taplo-common/src/schema/mod.rs
```diff
@@ -648,19 +651,18 @@ impl ValueExt for Value {
...
         &self,
-        _root_schema: &serde_json::Value,
-        url: &Url,
-        _original_ref: &str,
-    ) -> Result<Arc<serde_json::Value>, jsonschema::SchemaResolverError> {
+        uri: &jsonschema::Uri<&str>,
+    ) -> Result<Value, Box<dyn std::error::Error + Send + Sync>> {
         self.cache
-            .get_schema(url)
+            .get_schema(&Url::parse(uri.as_str())?)
```
uncenter, line 664: Since jsonschema now uses fluent_uri:Uri instead of url::Url, we convert the Uri back to Url to avoid refactoring other major areas.

crates/taplo-common/src/schema/mod.rs
```diff
@@ -134,52 +136,53 @@ impl<E: Environment> Schemas<E> {
...
+                    schema_path: err.schema_path,
+                })
+                .collect::<Vec<_>>();
+
+            // We check whether there were any external schema errors,
+            // and retrieve the schemas accordingly.
+            let mut external_schema_requests: FuturesUnordered<_> = errors
+                .iter()
+                .filter_map(|err| {
+                    if let ValidationErrorKind::Referencing(ReferencingError::Unretrievable {
```
uncenter, line 154: I believe this here `ValidationErrorKind::Referencing(ReferencingError::Unretrievable)` should be what we want to refer to external schema errors?

crates/taplo-common/src/schema/mod.rs
```diff
@@ -134,52 +136,53 @@ impl<E: Environment> Schemas<E> {
...
-                                    let value = self.load_schema(url).await?;
-                                    drop(self.cache.store(url.clone(), value));
-                                    Result::<(), anyhow::Error>::Ok(())
-                                })
-                            } else {
-                                None
-                            }
+            let errors = validator
+                .iter_errors(value)
+                .map(|err| ValidationError {
```
Stranger6667, line 141: Would `ValidationError::to_owned` work here? It does the same (except an extra clone that I've missed to remove)

crates/taplo-common/src/schema/mod.rs
Stranger6667: It is unfortunate that `referencing` does not support async resource retrieval, but yes, this error indicates that some resources can't be retrieved.

I'll prioritize supporting async retrieval to avoid re-building a validator as a way to deal with external schemas in an async manner

crates/taplo-common/src/schema/mod.rs
Stranger6667: You can safely drop this one - it was a not-so-useful optimization for the `LocationSegment` size. The idea was that paths in schemas use keys from a very limited set and I thought that instead of using a boxed str, it could be a string slice, hence `PathChunk::Keyword`. That approach would work better with a custom enum anyway (which is already there for some other parts)

Stranger6667 (approved):
Thank you! I forked `taplo` to implement it myself, but didn't manage to finish all the required changes there.

I've added a small note, but otherwise, everything looks great!

P.S. It is actually super useful to see how the new `jsonschema` versions are used in the real world. Specifically seeing what is done to handle the absence of async retrieval gives me a feeling that this feature should be higher on my todo list than now.

crates/taplo-common/src/schema/mod.rs
```diff
@@ -134,52 +136,48 @@ impl<E: Environment> Schemas<E> {
...
-                            if let ValidationErrorKind::Resolver { url, .. } = &err.kind {
-                                Some(async {
-                                    let value = self.load_schema(url).await?;
-                                    drop(self.cache.store(url.clone(), value));
-                                    Result::<(), anyhow::Error>::Ok(())
-                                })
-                            } else {
-                                None
-                            }
+            let errors = validator
```
Stranger6667, line 139: I realize now that in `0.28` all the external schemas retrieving happens during the validator compilation, not during validation, hence it this logic should be moved to `create_validator`

Stranger6667 (changes requested):
Left a comment about the overall validation flow, the current approach won't work

crates/taplo-common/src/schema/mod.rs
uncenter: Sorry, can you elaborate? How has the behavior changed from what this comment used to describe? Should I wait until async retrieval is implemented in jsonschema as you mentioned?

Stranger6667: Sure! `jsonschema` used to retrieve external documents on demand, during the `validate` call (or `is_valid` or `apply`) in order to perform the validation for `$ref` and related reference keywords. To make it work, the `$ref` implementation had some caching in place which implied a certain performance penalty (rwlock + inserts).

Some time ago I realized that it could be done just once when we create a validator (e.g. via `jsonschema::validator_for`) and all that caching won't be needed during validation as we'll have everything already retrieved, so I moved all that retrieval logic to the building stage.

I think it is not necessary to wait for async retrieval in `jsonschema`, as the loop approach would still work - it just needs to be applied inside `create_validator` (which requires making a few extra functions async here). If there are any external schemas missing, then the call to `ValidationOptions::build` will fail with the same validation errors as implemented in this PR

Stranger6667: I.e. the crucial change to this comment is that it is not about reaching some validation path, but rather discovering an external referencing during the building process.

Because all the resolving became eager and now happens much earlier in the process.

Here is also a plan I have for async retrieval:

- `0.29` will have `ValidationOptions::with_registry` that will expect a registry with all the needed schemas
- This `jsonschema::Registry` will have an async constructor that will accept resources and traverse them all until all the external schemas are retrieved and all the inner indices are built.

So, the retrieval will happen in batches where each batch will load previously unseen external resources that are reachable right now. Then the output of this batch will be traversed in the same way and if there are some new resources, there will be another batch, and so on.

Practically speaking:

```rust
let registry = jsonschema::Registry::new_async("http://127.0.0.1/dependent.json", json!({"$ref": "http://127.0.0.1/another.json"})).await?;

let validator = jsonschema::options()
    .with_registry(registry)
    .build({"$ref": "http://127.0.0.1/dependent.json"})?;
```

And maybe there could be just an async builder

```rust
let validator = jsonschema::non_blocking::options()
    .build({"$ref": "http://127.0.0.1/dependent.json"})
    .await?;

// Or

let validator = jsonschema::non_blocking::validator_for({"$ref": "http://127.0.0.1/dependent.json"})
    .await?;
```

uncenter: > Some time ago I realized that it could be done just once when we create a validator (e.g. via `jsonschema::validator_for`) and all that caching won't be needed during validation as we'll have everything already retrieved, so I moved all that retrieval logic to the building stage.

If I map over the errors from the initial validator build, will all of the retrieval errors surface at once or do I still need to continually loop and rebuild the validator until each and every remote reference is resolved?

Stranger6667: You still need to loop because external schemas can have references to other external schemas too and we don't know it until we load them.

Stranger6667: With `jsonschema==0.29` & the `resolve-async` feature enabled, the following should be enough on the validation build step:

```rust
jsonschema::async_options()
    .with_retriever(CacheSchemaRetriever {
        cache: self.cache().clone(),
    })
    .with_format("semver", formats::semver)
    .with_format("semver-requirement", formats::semver_req)
    .build(schema)
    .await
    .map_err(|err| anyhow!("invalid schema: {err}"))
```

+ retriever impl requires a few extra changes - using `async fn` and caching the retrieved value. 

Generally, the built-in retriever could work, but as I see here the cache is shared for all validators. I think in the future it could be done by having a shared `Registry` to avoid re-processing already processed external resources, but there is some extra work to be done on the `jsonschema` side. Right now I think the new `jsonschema` version should already simplify the logic quite a bit.

uncenter:
With the future of the project in a bit of jeopardy, I'm going to close this pull request for now. May revisit once project management is stabilized.

PR #732, opened 2025-01-11
Allow rules to target individual entries

### LINKED ISSUE #608: reorder_arrays = false is ignored if the key is set to the array.
I have the following `taplo.toml` file:
```toml
include = ["**/*.toml"]
[formatting]
reorder_arrays = true

[[rule]]
include = ["bar.toml"]
keys = ["baz"]

[rule.formatting]
reorder_arrays = false
```
I would assume that the rule listed would prevent the array `baz` from being re-ordered, but still allow for all other arrays to be sorted.

This is my `bar.toml` file:
```toml
baz = ["foo", "bar", "baz"]

sort_me = ["z", "a"]
```

However, when running `taplo fmt`, `bar.toml` ends up like this:
```toml
baz = ["bar", "baz", "foo"]

sort_me = ["a", "z"]
```

As you can see, all arrays were sorted. When omitting `keys`, nothing gets formatted, which is not what I want.

rmburg:
Fixes #608

### DIFF
--- a/crates/taplo/src/formatter/mod.rs
+++ b/crates/taplo/src/formatter/mod.rs
@@ -804,11 +804,15 @@ fn format_key(node: SyntaxNode, formatted: &mut String, _options: &Options, _con
 fn format_value(node: SyntaxNode, options: &Options, context: &Context) -> impl FormattedItem {
     let mut value = String::new();
     let mut comment = None;
+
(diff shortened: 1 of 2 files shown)


rmburg:
Hi, is there anything keeping this from being merged?

panekj:
Yes, I need to review it first.

PR #743, opened 2025-02-10
ci: Build wheel for linux aarch64

phuongfi91:
Atm there's no wheel built for linux aarch64, this should address that

### DIFF
--- a/.github/workflows/releases.yaml
+++ b/.github/workflows/releases.yaml
@@ -458,6 +458,11 @@ jobs:
             arch: x86
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


JounQin:
How can we test it before merging?

JounQin:
How can we test it before merging?

panekj:
https://github.com/tamasfe/taplo/actions/runs/14724623952/job/41324620377?pr=743

panekj:
> this is blocking us from using taplo on aarch64 platform

this will have to wait a bit until I have my main PC back so I can prepare do to releases (hopefully this week I will be able to move my things from previous home)

PR #751, opened 2025-02-21
Support "2024" edition of Rust.

chaten:
The "2024" edition of Rust releases with Rust 1.85.
This updates the schema for Cargo.toml to allow for projects using this edition.


JounQin (approved):
You'd better provide related link describing it.

panekj:
It's already present in the schema: https://github.com/SchemaStore/schemastore/blob/088962593fd3274b0ae1872688951e6e4edf2f42/src/schemas/json/cargo.json#L245

chaten:
@panekj Apologies, I'm very unfamiliar with both this project and that one. However, that looks like a very different file than the one modified here. The file modified here is being downloaded by a Vim plugin [coc-toml](https://github.com/kkiyama117/coc-toml) via "https://taplo.tamasfe.dev/schemas/Cargo.toml.json", which as of today will error out on a new project's Cargo.toml.
Coc-toml is most likely an abandoned project from a newer dev, so maybe they used these files when they weren't supposed to be relied on?

@JounQin Rust 1.85 released today with support for 2024, https://doc.rust-lang.org/edition-guide/rust-2024/index.html , https://releases.rs/docs/1.85.0/

panekj:
> Coc-toml is most likely an abandoned project from a newer dev, so maybe they used these files when they weren't supposed to be r

see https://github.com/kkiyama117/coc-toml/issues/115

PR #756, opened 2025-03-05
Add support for Rust edition 2024

jeff-hiner:
Currently supported as of Rust stable 1.85.0, so we should also make sure setting `edition = "2024"` doesn't throw errors.


panekj:
Duplicate of #751

jeff-hiner:
@panekj Was there a reason https://github.com/tamasfe/taplo/pull/751 was closed without merging? This issue is currently causing the plugin to fail currently on any 2024 edition Cargo.toml.

panekj:
> Was there a reason #751 was closed without merging?

The reason is stated in the PR.



> Is the pull from schemastore just not yet released?

It has been available since March 29, 2024
https://github.com/SchemaStore/schemastore/commit/6e4146f660766256fbd2999ef37366fa57a508c2

jeff-hiner:
Okay. So do I have to do something special within vscode to get my valid Cargo.toml to not throw errors?

panekj:
Check what schema is used when you have the file open and try switching it manually as well
image

panekj:
It should be `Cargo Manifest`

jeff-hiner:
It was already Cargo Manifest, but for some reason it was using an outdated version of the cargo.json from upstream. Deleting the `~/.config/Code/User/globalStorage/tamasfe.even-better-toml/` contents and restarting vscode fixed it, but there's probably some sort of caching bug still.

PR #774, opened 2025-04-12
support multiline and trailing comma in inline tables

### LINKED ISSUE #333: Line breaks and trailing comma in inline tables
Line breaks and trailing commas are properly parsed but will emit a syntax error. In order to support https://github.com/toml-lang/toml/pull/904 we could drop these errors.

vemoo:
fixes #333 

Initially I deleted some files from `test-data/invalid` and tried to regenerate `crates/taplo/src/tests/generated/invalid.rs` by running `cargo run --manifest-path ./util/test-gen/Cargo.toml -- -i test-data -o crates/taplo/src/tests/` but it doesn't match the existing file, so I edited `crates/taplo/src/tests/generated/invalid.rs` by hand.

It may be a good idea consider using https://github.com/toml-lang/toml-test to test the parser.

### DIFF
--- a/crates/taplo/src/parser/mod.rs
+++ b/crates/taplo/src/parser/mod.rs
@@ -806,6 +806,9 @@ impl<'p> Parser<'p> {
                     self.add_token()?;
                 }
                 COMMA => {
+                    if !expect_comma_or_end {
+                        let _ = self.error(r#"unexpected ",""#);
+                    }
                     self.add_token()?;
                     expect_comma_or_end = false;
                 }


vemoo:
> It may be a good idea consider using https://github.com/toml-lang/toml-test to test the parser.

😅. I see it's used on CI. I was just testing with `cargo test`. Will take a look at the `toml-test` failures.

vemoo:
The issue was that toml-test 1.1.0 only supports TOML 1.0. I have updated the CI workflow to use 1.5.0 and TOML 1.1.0 but the issue now is that there are other tests failing not related to the changes on this PR.

Those tests should probably be fixed in a separate PR. They can be skipped when invoking toml-test, and un-skip them in later PR or leave them failing for now.

panekj:
Fixed via #849

PR #786, opened 2025-05-09
clear vscode extensions

CoffeeChaton:
image

This appears to be a development-only script, not needed at runtime. If relying on a blacklist makes it easy to miss things, a whitelisting approach might be more reliable. For a similar concept, see the discussion here: <https://github.com/microsoft/vscode-vsce/issues/12>.

### DIFF
--- a/editors/vscode/.vscodeignore
+++ b/editors/vscode/.vscodeignore
@@ -16,3 +16,7 @@ node_modules
 .yarnrc.yml
 rollup-config.js
 scripts/
+node.tsconfig.json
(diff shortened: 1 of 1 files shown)


JounQin (approved):

editors/vscode/.vscodeignore
```diff
@@ -16,3 +16,7 @@ node_modules
 .yarnrc.yml
 rollup-config.js
```
JounQin, line 17:
```suggestion
```

JounQin (changes requested):

PR #789, opened 2025-05-12
fix: remove all the array index keys when there is more than one

### LINKED ISSUE #788: Hover not working on keys whose value is a non-empty array
When using a JSON schema, hovering over a key triggers the LSP "hover" action, which displays the description of that key. However, I noticed that this doesn't work when the value of that key is a non-empty array.

Sample TOML file to reproduce the issue (`hover.toml`):

```toml
# Not an array value: hover on the key works fine
string = "foo"
# Empty array value: hover on the key works fine
empty_array = []
# Non-empty array value: hover on the key is broken
non_empty_array = ["bar"]
```

Corresponding JSON schema:

```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "title": "Hover Test Schema",
    "description": "A schema to test hover functionality in Taplo",
    "type": "object",
    "properties": {
        "string": {
            "type": "string",
            "description": "A simple string value. Hover on this key should work fine."
        },
        "empty_array": {
            "type": "array",
            "description": "An empty array. Hover on this key should work fine.",
            "items": {
                "type": "string"
            }
        },
        "non_empty_array": {
            "type": "array",
            "description": "A non-empty array. If you see this description when hovering, it means the bug is fixed.",
            "items": {
                "type": "string"
            }
        }
    }
}
```

I used the following `~/.taplo.toml` file to connect the two (adjust the path to the schema accordingly):

```toml
[[rule]]
include = ["**/hover.toml"]

[rule.schema]
path = "/path/to/hover-schema.json"
```

ipkiss42:
The `Query::position_info_at()` function can return one index when the value is a non-empty array. In this case, the `Query::lookup_keys()` function will add another one, resulting in two index keys. But the code was only removing one, causing hover to break for keys with a non-empty array value.

This change avoids the issue by removing index keys in a while loop.

Closes #788

### DIFF
--- a/crates/taplo-lsp/src/handlers/hover.rs
+++ b/crates/taplo-lsp/src/handlers/hover.rs
@@ -112,7 +112,7 @@ pub(crate) async fn hover<E: Environment>(
             keys = lookup_keys(doc.dom.clone(), &keys);
 
             // We're interested in the array itself, not its item type.
-            if let Some(KeyOrIndex::Index(_)) = keys.iter().last() {
+            while let Some(KeyOrIndex::Index(_)) = keys.iter().last() {
                 keys = keys.skip_right(1);
             }
 


panekj:
Merged in https://github.com/tamasfe/taplo/commit/db54f8de3269ba9093821c41e31d890e663ee523

Issue #470, opened 2023-09-12
Github Action / running taplo in CI


jbergstroem:
It's easy enough to get taplo running in CI. Something like this should work (and version can be maintained with a regex in renovatebot):

```yaml
  taplo:
    name: Taplo
    runs-on: ubuntu-22.04
    steps:
      - name: Checkout
        uses: actions/checkout@v4
      - name: Install Taplo
        env:
          version: "0.8.1"
        run: |
          curl -Ls "https://github.com/tamasfe/taplo/releases/download/${{ env.version }}/taplo-full-linux-x86_64.gz" | \
          gzip -d > taplo && \
          chmod +x taplo && \
          sudo mv taplo /usr/local/bin/taplo
      - name: Run Taplo
        run: taplo validate --colors never --no-schema
```

..that said. The very helpful CLI output is very complicated to parse with a [problem matcher](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md).

One way to improve this could be to support a custom output template, [similar to actionlint](https://github.com/rhysd/actionlint/blob/main/docs/usage.md#format-error-messages). Another approach could be to have a custom format specifically catering problem matchers (which also could serve as a way get improved editor support).

ia0:
Thanks for the suggestion!

I'm personally not convinced by sophisticated output (others may disagree). I just use Taplo in CI like [this](https://github.com/google/wasefire/blob/59f73b90dd0c7e3b232976a37b055fe506360c6e/scripts/ci.sh#L63-L64), i.e. by formatting then diffing using git (which has a nice unsophisticated output). (This is also related to #416 when CI as a use-case for Taplo was mentioned.)

That said, I'm not against this feature if someone is ready to design and implement it and someone else is ready to review it.

panekj:
I think we could output https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message when `GITHUB_CI` is set (or whichever envvar GitHub used).

jbergstroem:
Was just coming back to suggest that! Probably the best approach.

> when GITHUB_CI is set (or whichever envvar GitHub used).

Thinking either `GITHUB_ACTIONS` on its own or together with `CI`? 
https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

panekj:
> Thinking either `GITHUB_ACTIONS` on its own

This should be sufficient.

uncenter:
[...]

PR #790, opened 2025-05-17
feat(action): add github action for cargo.toml formating

gwen-lg:
_Note: this is a first version for comments_

It's great to have Cargo.toml (by example) formatted on save in VSCode. But If it's not checked in CI, this is dependent of developer use `Even Better TOML` and process.

TODO:
- manage config with `indent_string="    "` for Cargo.toml (like `Even Better TOML`)
- convert output to display beautiful error/warning in github
- feature for allow automatic patch commits ?

#470

Work inspired from https://github.com/marketplace/actions/typos-action

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,0 +1,36 @@
+name: "cargo.toml fmt"
+author: "Gwen Lg"
+description: "Run TOML formatting change or check on Cargo.toml in GitHub Actions"
+
+inputs:
+  version:
+    description: "Version of Taplo"
+    required: false
+    default: "0.9.3"
+
+  files:
+    description: "Files or patterns to check"
+    required: false
+    default: "Cargo.toml"
+
+  write_changes:
+    description: "Do the formatting and write changes to the repository"
+    required: false
+    default: false
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Install Taplo
+      id: fmt
+      shell: bash
+      env:
+        INSTALL_DIR: .
+        INPUT_VERSION: ${{ inputs.version }}
+        INPUT_FILES: ${{ inputs.files }}
+        INPUT_WRITE_CHANGES: ${{ inputs.write_changes }}
+      run: $GITHUB_ACTION_PATH/action/entrypoint.sh
+
+branding:
+  icon: "type"
+  color: "red"
--- a/action/entrypoint.sh
+++ b/action/entrypoint.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+set -eu
+
+SOURCE_DIR="$(dirname -- ${BASH_SOURCE[0]:-$0})"
+
+log() {
+    echo -e "$1" >&2
+}
+
+_DEFAULT_INSTALL_DIR=${HOME}/bin
+_INSTALL_DIR=${INSTALL_DIR:-${_DEFAULT_INSTALL_DIR}}
+CMD_NAME="taplo"
+COMMAND="${_INSTALL_DIR}/${CMD_NAME}"
+
+TARGET=${INPUT_FILES:-"."}
+if [[ -z $(ls ${TARGET} 2>/dev/null) ]]; then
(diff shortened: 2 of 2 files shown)


action.yml
```diff
@@ -0,0 +1,36 @@
...
+inputs:
+  version:
+    description: "Version of Taplo"
+    required: false
+    default: "0.9.3"
+
+  files:
+    description: "Files or patterns to check"
+    required: false
+    default: "Cargo.toml"
```
tobiasdiez, line 14: I think it would be more universally useful to make the files required and don't specify a default.

action/entrypoint.sh
```diff
@@ -0,0 +1,63 @@
...
+        unzip -o "${FILE_NAME}" -d ${_INSTALL_DIR} ${CMD_NAME}.exe
+    else
+        gzip -d "${FILE_NAME}" -c >${COMMAND}
+        chmod +x ${COMMAND}
+    fi
+    rm "${FILE_NAME}"
+fi
+log "jq: $(jq --version)"
+
+ARGS="format"
```
tobiasdiez, line 55: Would be great to also have support for `taplo check`. And perhaps rename `format` to `fmt` (at least the docs seem to prefer fmt).

tobiasdiez (commented):
Thanks, great initiative! A few suggestions from a cursory read of the diff.

panekj:
Taplo is already usable in GHA via `uses: docker://tamasfe/taplo`

panekj:
> Why not [autofix.ci](https://autofix.ci)?

Maybe someone doesn't want to pay for it?

JounQin:
> Maybe someone doesn't want to pay for it?

Hah, isn't it free?

panekj:
> > Maybe someone doesn't want to pay for it?
> 
> Hah, isn't it free?

Not for private repos

JounQin:
> Not for private repos

How is it related to private repos here?

panekj:
You brought up autofix.ci?

JounQin:
> You brought up autofix.ci?

But taplo is not a private repo?

JounQin:
> And the PR isn't about using taplo for taplo repository. You're suggesting using autofix.ci for something, I don't know what. The PR adds GitHub actions for uing taplo in GHA which can be used in public repos, private repos, GHES, whatever.

Sorry I misunderstood what this PR does. :(

action.yml
```diff
@@ -0,0 +1,36 @@
...
+  steps:
+    - name: Install Taplo
+      id: fmt
+      shell: bash
+      env:
+        INSTALL_DIR: .
+        INPUT_VERSION: ${{ inputs.version }}
+        INPUT_FILES: ${{ inputs.files }}
+        INPUT_WRITE_CHANGES: ${{ inputs.write_changes }}
+      run: $GITHUB_ACTION_PATH/action/entrypoint.sh
```
JounQin, line 32: Does it work for Windows?

action/entrypoint.sh
gwen-lg: as `fmt` is an alias for `format`, it's seems most logical to me to use `format` unless `fmt` presented as preferred way ?
I haven't explored the `check`/`lint` functionality, what it the interest in CI ? What you think it's should do ? How should it be integrated ?

tobiasdiez: I cannot say anything about the preference of `fmt` vs `format` except that the docs seem to only mention `fmt`

The check would work similarly as for `fmt`: run `taplo check`, if it reports a problem fail. So my proposal would be to add two boolean toggles (both defaulting to true?), each for enabling/disabling the the fmt and check run.

gwen-lg: Ok, thank.

tobiasdiez:
Docker-based actions are not available for macos and windows (based on https://docs.github.com/en/actions/sharing-automations/creating-actions/about-custom-actions#types-of-actions)

PR #794, opened 2025-05-23
Reduce logging verbosity

3tilley:
We have a lot of toml files, and by default the CLI emits at INFO level. I would also be happy with a complete approach, e.g. adding a --quiet flag, or only emitting logs at warning level, but this I think is a generally good change.


panekj:
Merged in https://github.com/tamasfe/taplo/commit/b673b44df2773db8673a00df2e7654b769f7fde7

PR #814, opened 2025-08-30
Update lock file to avoid GHSA-xwfj-jgwm-7wp5

jayvdb:
See https://osv.dev/GHSA-xwfj-jgwm-7wp5


jayvdb:
@panekj In order to have this reviewed, and hopefully merged.
cc @mrexodia

PR #828, opened 2025-11-07
Updates schemastore endpoint

konn:
It seems that Schema Store recently changes their API endpoint to www.schemastore.org, instead of old json.schemastore.org for some reason. For example, although I could not find any actual discussion or announcement, this is stated here: https://github.com/microsoft/vscode/issues/254689.

It seems that json.* subdomain had lived for some duration, but it starts to fail to serve reliably, and results in unexpected linting error.
This PR aims at fixing this just by rewriting schemastore catalog url. I confirmed that this should work by adding following lines to `.taplo.toml`:

```toml
[schema]
path = "https://www.schemastore.org/api/json/catalog.json"
```

### DIFF
--- a/crates/taplo-common/src/schema/associations.rs
+++ b/crates/taplo-common/src/schema/associations.rs
@@ -504,7 +504,7 @@ pub struct SchemaStoreSchemaMeta {
 }
 
 pub const SCHEMA_STORE_CATALOG_SCHEMA_URL: &str =
-    "https://json.schemastore.org/schema-catalog.json";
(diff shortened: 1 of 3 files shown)


konn:
@denysvitali-kaiko Oh, I didn't notice that. Thank you for your note! Meanwhile, I let this PR just as-is.

AlexTMjugador:
The change proposed by this PR has been effectively applied as of https://github.com/tamasfe/taplo/commit/08f343be02ce1b20296470396a42f0fa47820449 :slightly_smiling_face:

Issue #715, opened 2024-12-20
The future of the project


tamasfe:
**tl;dr**: I can't see myself returning here, and I'm asking you to decide the fate of this project.

When I started working on taplo, I wanted it to be a tool that puts TOML on par with YAML and JSON in terms of dev experience and perhaps even surpass them, in some aspects I think I achieved that and I'm happy that it ended up being useful for many. However the project was pretty much a learning experience for me, it was the first parser I have written, the first LSP, etc. and it really shows.

I still have a lot of plans, for example:

- rework the parser so that it is not a mess (e.g. I really liked the approach I ended up with in [rhai LSP](https://github.com/rhaiscript/lsp/blob/master/crates/rhai-rowan/src/parser/parsers/mod.rs), I think `rust-analyzer` has a similar architecture)
- adjust the code that builds on top of it so that it won't rely on an [accidentally introduced behavior that actually makes no sense](https://github.com/tamasfe/taplo/blob/a8bc571ee28775e7d5ad84c2ea87cf5b61ab42f5/crates/taplo/src/parser/mod.rs#L185)
- rethink/rework whatever the hell is happening after syntax parsing, to date I still have no idea how to efficiently and cleanly turn TOML into a DOM with its arbitrary rules and parent nodes appearing out of order
- rebase the LSP onto a maintained library, like [tower-lsp](https://github.com/ebkalderon/tower-lsp)
- rework the formatter so that it handles line breaks and nesting properly, something like [prettyplease](https://github.com/dtolnay/prettyplease) or [rhai-fmt](https://github.com/rhaiscript/lsp/tree/master/crates/rhai-fmt) (that was also based on prettyplease).
- fix filepath/URL handling, I think there is still confusion around this (unless the `windows<->everything else` conflicts were fixed while I was away)
- schemas and in general position querying (which TOML node maps to which property in schemas) is quite buggy
- ... probably many more (looking at the open issues)

I have practically disappeared from this project for the past 1-2 years, and now it is safe to say that whatever my plans are, they are very unlikely to be fulfilled. Nowadays I have a lot less free time, and even in that free time I code less as a hobby. On top of that being a maintainer really feels like a 2nd or 3rd job and I have a lot of respect for people who can pull it off, I definitely cannot.

On top of that, I don't feel like that the amount of work I listed above (probably a month worth?) is worth putting into this tool, while it is useful, I don't feel anyone would really miss it if it disappeared overnight.

[...]

PR #844, opened 2026-02-10
fix(lsp): include request id in error responses

graelo:
Hello, thanks for taplo! I use it quite often. I was annoyed by the following, so here's a PR with a fix.

Best regards!

## Problem

When Neovim (or any LSP client) sends a `shutdown` request to taplo, the exit is
delayed by ~1 second. The client logs:

```
LSP[taplo]: Error NO_RESULT_CALLBACK_FOUND: {
  error = {
    code = -32600,
    data = "server is shutting down",
    message = "Invalid request"
  },
  id = 0,
  jsonrpc = "2.0"
}
```

## Root cause

In `crates/lsp-async-stub/src/lib.rs`, two error response paths omit
`.with_request_id(...)`, so the response is sent with the hardcoded
default `id: 0` (from `Response::error()` in `rpc.rs`):

1. **"server is shutting down"** (line 381-388) — when a request arrives
   after `shutting_down` is set to `true`
2. **"server not initialized"** (line 412) — when a request arrives
   before initialization

The client cannot match a response with `id: 0` to any pending request,
so it ignores it and waits for a timeout before force-killing the server.

Other error paths in the same function (`method_not_found` at line 483,
shutdown success at line 470) correctly call `.with_request_id()`.

## Fix

Add `.with_request_id(request.id.clone().unwrap())` to both error
response paths, matching the existing pattern.

## Reproduction

```bash
# Create a minimal TOML file
echo '[package]\nname = "test"' > /tmp/test.toml

# Baseline: open a non-TOML file (~0.15s overhead)
time nvim --headless -c "edit /tmp/test.lua" -c "sleep 1" -c "qall" 2>&1

# Bug: open a TOML file (~1.1s overhead due to taplo shutdown delay)
time nvim --headless -c "edit /tmp/test.toml" -c "sleep 1" -c "qall" 2>&1
```

Before fix: the TOML case takes ~2.1s (1s sleep + 1.1s shutdown delay).
After fix: the TOML case should take ~1.15s (1s sleep + ~0.15s overhead),
matching the non-TOML baseline.

## Related

- PR #354 — fixed taplo not responding to shutdown *at all*
- Issue #400 — duplicate of #354

### DIFF
--- a/crates/lsp-async-stub/src/lib.rs
+++ b/crates/lsp-async-stub/src/lib.rs
@@ -383,6 +383,7 @@ impl<W: Clone> Server<W> {
                         rpc::Response::error(
                             rpc::Error::invalid_request().with_data("server is shutting down"),
                         )
+                        .with_request_id(request.id.clone().unwrap())
                         .into_message(),
                     )
                     .await?;
@@ -409,7 +410,11 @@ impl<W: Clone> Server<W> {
                 );
 
                 writer
-                    .send(rpc::Response::error(rpc::Error::server_not_initialized()).into_message())
+                    .send(
+                        rpc::Response::error(rpc::Error::server_not_initialized())
+                            .with_request_id(request.id.clone().unwrap())
+                            .into_message(),
+                    )
                     .await?;
                 return Ok(());
             }


panekj:
I would urge people to just use whatever is the latest commit on the branch, see #715 for more context

Issue #799, opened 2025-06-04
error: data did not match any variant of untagged enum SchemaCatalog


serl:
Hello, since a couple of hours, all of our `taplo-lint` pre-commit checks are failing.

Seems like the [default hook](https://github.com/ComPWA/taplo-pre-commit/blob/main/.pre-commit-hooks.yaml) is passing `--default-schema-catalogs` and that fails:

```
$ uvx taplo lint --default-schema-catalogs
 INFO taplo:load_config: found configuration file path="[...]/.taplo.toml"
 WARN taplo: failed to fetch catalog error=error decoding response body: data did not match any variant of untagged enum SchemaCatalog

Caused by:
    data did not match any variant of untagged enum SchemaCatalog
ERROR operation failed error=failed to load schema catalog: error decoding response body: data did not match any variant of untagged enum SchemaCatalog: data did not match any variant of untagged enum SchemaCatalog
```

I guess that's linked to the latest commit in https://github.com/schemastore/schemastore, 4 hours ago.

wadhah101:
having this too

JounQin:
Thanks @hyperupcall! Closing in favor of https://github.com/SchemaStore/schemastore/pull/4761

palfrey:
This one is back!
```
     Running `target/debug/taplo lint --default-schema-catalogs`
 INFO taplo:load_config: found configuration file path="/home/palfrey/src/taplo/taplo.toml"
 WARN taplo: failed to fetch catalog error=error decoding response body: data did not match any variant of untagged enum SchemaCatalog

Caused by:
    data did not match any variant of untagged enum SchemaCatalog
ERROR operation failed error=failed to load schema catalog: error decoding response body: data did not match any variant of untagged enum SchemaCatalog: data did not match any variant of untagged enum SchemaCatalog
```
With current master

hamza-56:
Also hitting this in pre-commit (via ComPWA/taplo-pre-commit#35). Reproduces with both `taplo 0.9.3` (the latest in any pre-commit mirror) and a directly-installed `taplo 0.10.0` when `--default-schema-catalogs` is passed:

```
$ taplo lint --default-schema-catalogs pyproject.toml
 WARN taplo: failed to fetch catalog error=error decoding response body: data did not match any variant of untagged enum SchemaCatalog
ERROR operation failed error=failed to load schema catalog
```

[...]

Issue #463, opened 2023-08-22
Schemas not working in Neovim


MariaSolOs:
I'm not getting any kind of schema validation in Neovim. I'm using [`nvim-lspconfig`](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#taplo) to setup the server, and the logs shows the following when opening a simple `Cargo.toml` file:
```
[START][2023-08-21 19:44:24] LSP logging initiated
[ERROR][2023-08-21 19:44:24] .../vim/lsp/rpc.lua:675	"rpc"	"/Users/majosolano/.local/share/nvim/mason/bin/taplo"	"stderr" 
INFO taplo: registered request handler method="initialize"\n 
INFO taplo: registered request handler method="textDocument/foldingRange"\n 
INFO taplo: registered request handler method="textDocument/documentSymbol"\n 
INFO taplo: registered request handler method="textDocument/formatting"\n 
INFO taplo: registered request handler method="textDocument/completion"\n
INFO taplo: registered request handler method="textDocument/hover"\n
INFO taplo: registered request handler method="textDocument/documentLink"\n
INFO taplo: registered request handler method="textDocument/semanticTokens/full"\n
INFO taplo: registered request handler method="textDocument/prepareRename"\n
INFO taplo: registered request handler method="textDocument/rename"\n
INFO taplo: registered notification handler method="initialized"\n
INFO taplo: registered notification handler method="textDocument/didOpen"\n
INFO taplo: registered notification handler method="textDocument/didChange"\n
INFO taplo: registered notification handler method="textDocument/didSave"\n
INFO taplo: registered notification handler method="textDocument/didClose"\n
INFO taplo: registered notification handler method="workspace/didChangeWorkspaceFolders"\n
INFO taplo: registered request handler method="taplo/convertToJson"\n
INFO taplo: registered request handler method="taplo/convertToToml"\n
INFO taplo: registered notification handler method="workspace/didChangeConfiguration"\n
INFO taplo: registered request handler method="taplo/listSchemas"\n
INFO taplo: registered request handler method="taplo/associatedSchema"\n
INFO taplo: registered notification handler method="taplo/associateSchema"\n
INFO taplo: LSP server listening transport="stdio"\n'
[ERROR][2023-08-21 19:44:25] .../vim/lsp/rpc.lua:675	"rpc"	"/Users/majosolano/.local/share/nvim/mason/bin/taplo"	"stderr"	"
[...]

PR #866, opened 2026-05-05
Replace json.schemastore.org with www.schemastore.org

palfrey:
I've just hit https://github.com/tamasfe/taplo/issues/799 again which I believe is related to https://github.com/SchemaStore/schemastore/issues/5621.

```
 Running `target/debug/taplo lint --default-schema-catalogs`
 INFO taplo:load_config: found configuration file path="/home/palfrey/src/taplo/taplo.toml"
 WARN taplo: failed to fetch catalog error=error decoding response body: data did not match any variant of untagged enum SchemaCatalog

Caused by:
    data did not match any variant of untagged enum SchemaCatalog
ERROR operation failed error=failed to load schema catalog: error decoding response body: data did not match any variant of untagged enum SchemaCatalog: data did not match any variant of untagged enum SchemaCatalog
```

[Advice from there](https://github.com/SchemaStore/schemastore/issues/5621#issuecomment-4329392486) was "Btw, everyone should use [www](http://www/). instead of json. subdomains for accessing anything on schemastore.org. If you have any hardcoded values pointing to json., then replace it now with www." which is what this PR does and it makes things work again


palfrey:
https://github.com/tamasfe/taplo/pull/828 does most of this as well

AlexTMjugador:
The change proposed by this PR has been effectively applied as-is as of https://github.com/tamasfe/taplo/commit/08f343be02ce1b20296470396a42f0fa47820449.

AlexTMjugador:
At the moment, `taplo lint --schema-catalog https://www.schemastore.org/api/json/catalog.json Cargo.toml` is still not working for me when linting a Rust project file with the following error:

```
 WARN taplo: failed to fetch catalog error=error decoding response body: data did not match any variant of untagged enum SchemaCatalog

Caused by:
    data did not match any variant of untagged enum SchemaCatalog
ERROR operation failed error=failed to load schema catalog: error decoding response body: data did not match any variant of untagged enum SchemaCatalog: data did n
ot match any variant of untagged enum SchemaCatalog
```

However, visiting https://www.schemastore.org/api/json/catalog.json in a web browser returns a valid-looking JSON response.

The older schema catalog suggested in https://github.com/tamasfe/taplo/issues/463#issuecomment-1689244621 is no longer working for me either.

It seems likely that `taplo` needs further tweaks to handle a change in the schema catalog format.

PR #882, opened 2026-08-18
fix(vscode): ship SchemaStore fixes in the bundled LSP

### LINKED ISSUE #799: error: data did not match any variant of untagged enum SchemaCatalog

joshmouch:
## Why this change exists

SchemaStore catalog loading can fail with `data did not match any variant of untagged enum SchemaCatalog` when the catalog's `$schema` URL moves between its legacy, current, and raw-GitHub identifiers ([#799](https://github.com/tamasfe/taplo/issues/799)). Updating the catalog endpoint alone leaves that exact-identifier failure mode in place.

Even Better TOML also still bundles the published `@taplo/lsp@0.8.0` package. As a result, fixes present in this repository do not reach a newly packaged extension until a separate npm release happens. This PR makes SchemaStore catalog identification independent of the URL host and builds the extension's bundled LSP from the same source checkout.

## What changes

- accept SchemaStore `$schema` URLs whose path identifies `schema-catalog.json`, while preserving Taplo-format catalog detection and the existing public Rust API
- build and test `@taplo/lsp` from this repository before bundling either VS Code server
- use the current wasm-bindgen initialization shape and JSON-compatible LSP serialization
- avoid unsupported-handler warnings for the standard `$/setTrace` notification
- make the build sequential and failure-propagating, install the WebAssembly target in release CI, and use Yarn 4's npm publication lifecycle and command
- add regressions for the observed SchemaStore identifiers, Taplo-catalog discrimination, built-package JSON-RPC, and trace-notification handling

`--locked` is intentional: fresh release runners may fetch dependencies, while `Cargo.lock` still fixes their versions. `--frozen` also forbids network access and fails on an uncached runner.

The JSON-compatible serializer overlaps with [#860](https://github.com/tamasfe/taplo/pull/860). It is included here because the source-built LSP needs the same transport correction; that PR's hover, completion, and other LSP changes remain independent.

This prepares both Even Better TOML 0.21.3 and `@taplo/lsp` 0.8.2. The extension no longer has to wait for the npm package to be published before it can ship the repository's current LSP.

Fixes #799.

## Verification

On the current review revision:

- `cargo check -p taplo-common --tests --features schema,reqwest,rustls-tls`
- focused tests for all three SchemaStore identifiers, invalid identifiers, and Taplo-format discrimination
- `cargo test -p taplo-lsp`
- `cargo check -p taplo-wasm --target wasm32-unknown-unknown`
- immutable Yarn workspace install and `@taplo/core` build
- JavaScript syntax checks and a failure-propagation check for the shared build helper

The pull-request workflows exercise the release build, built-package JSON-RPC regression, extension bundles, and VSIX packaging for the pushed head.

## Review focus

The main design choice is whether the extension should build the LSP from this checkout, as proposed here, or continue to depend on a separately published npm package. Marketplace publication access remains operationally separate from the source and packaging fixes in this PR.


joshmouch:
Could @panekj or @JounQin take a look at #882?

SchemaStore catalog loading can fail again when the catalog's `$schema` identifier moves between its legacy, current, and raw-GitHub URLs. Changing only the fetch endpoint does not remove that exact-identifier dependency. On upstream `master`, the VS Code extension still consumes the published `@taplo/lsp@0.8.0`, so repository fixes do not reach a newly packaged extension until a separate npm release happens.

The PR now accepts the observed catalog identifiers without changing Taplo-format catalog detection or the public Rust API, builds the bundled LSP from the checkout, and exercises both `yarn pack` and VSIX packaging in CI. Its JSON-compatible WASM serialization overlaps with #860; the hover, completion, and other LSP changes there remain independent.

The focused design question is whether the extension should build the LSP directly from the release checkout, as proposed here, or continue to depend on a separately published npm package. Marketplace publication access is operationally separate, but review of the source and packaging path would still unblock the code decision.

panekj:
I have a strong feeling that this was generated with LLM, in which case I will not even look at it and never accept it.

